PR #38, opened 2026-04-22
feat(cli): opt-in auth resolution for custom commands and dispatchers (#37)

### LINKED ISSUE #37: Custom commands: opt-in auth resolution and session persistence API
## Summary

Consumer-registered custom commands (`.apijack/commands/*.ts`) and dispatchers (`.apijack/dispatchers/*.ts`) receive a `CliContext` whose `session` field is lazily resolved. Generated OpenAPI commands trigger resolution automatically via `client.ensureReady()` inside the `ApiClient` wrapper (`cli-builder.ts:319`), but consumer code gets a raw `ctx` where `session` is `null` until something explicitly resolves it. There is no public API to signal that a custom command needs auth, and no public API to persist a mutated session back to disk.

This creates two classes of latent bugs that any non-trivial consumer hits:

1. **Null-deref** — custom code reading `ctx.session.*` before first resolution crashes with `TypeError: null is not an object`.
2. **Lost session mutations** — custom code assigning `ctx.session = newSession` has no way to persist it; `SessionManager` is not exported, and its `save()` method is private (`src/session.ts:65`). The next invocation loads a stale cache.

The framework currently expects consumers to know they must call an undocumented method on an `unknown`-typed `ctx.client` field. That contract is invisible from the type system, the types (`session: AuthSession | null`), and the current docs.

### 1. Module-level `requiresAuth` flag on commands and dispatchers

Consumers declare auth need alongside the existing `name` export:

```ts
// .apijack/commands/foo.ts
export const name = "foo";
export const requiresAuth = true;
export default function register(program, ctx) {
  program.command("foo").action(async () => {
    // ctx.session is guaranteed non-null here
  });
}
```

```ts
// .apijack/dispatchers/bar.ts
export const name = "bar";
export const requiresAuth = true;
export default async function handler(args, positional, ctx) { ... };
```

Core behavior:
- `loadProjectCommands` / `loadProjectDispatchers` (in `src/project-loader.ts`) read the new export and attach it to the loaded record.
- For commands: in the registration loop (`cli-builder.ts:406`), wrap the Commander subcommand with `.hook('preAction', async () => { await ctx.client.ensureReady(); })` when `requiresAuth` is true. Using `preAction` leaves the consumer's `.action()` registration untouched.
- For dispatchers: in `buildDispatcher` (`cli-builder.ts:435`), `await ctx.client.ensureReady()` before invoking the handler when `requiresAuth` is true.

### 2. Configurable default via `settings.json`

Default is `false` (backward-compatible). A consumer project can flip the default via a new framework-settings file `.apijack/settings.json` (sibling of `config.json`, separate from per-environment credentials):

```json
{
  "customCommands": {
    "defaults": {
      "requiresAuth": true
    }
  }
}
```

Resolution order for any given command/dispatcher: module-level `export const requiresAuth` overrides `settings.customCommands.defaults.requiresAuth` overrides built-in default (`false`).

Setup, config, routine, generate, plugin, and mcp built-ins are registered directly (`cli-builder.ts:154-171`), not via the project loader, and are unaffected.

### 3. New `CliContext` methods

```ts
interface CliContext {
  client: unknown;
  session: AuthSession | null;
  auth: ResolvedAuth;
  strategy: AuthStrategy;
  refreshSession(): Promise<void>;
  resolveSession(): Promise<void>;   // new — public alias for client.ensureReady()
  saveSession(): Promise<void>;      // new — persists ctx.session via SessionManager
}
```

- `resolveSession()` lets a consumer opt a single code path into resolution without the module-level flag (useful for one-off helpers).
- `saveSession()` closes the persistence gap. It writes `ctx.session` to disk using the same `SessionManager` instance already wired into the CLI; consumers never touch `SessionManager` directly.

### 4. Narrowed context type when auth is required

```ts
export interface CliContext {
  session: AuthSession | null;
  // ...
}

export interface AuthedCliContext extends CliContext {
  session: AuthSession;   // non-null
}

export type CommandRegistrar<R extends boolean = false> =
  R extends true
    ? (program: Command, ctx: AuthedCliContext) => void
    : (program: Command, ctx: CliContext) => void;

export type DispatcherHandler<R extends boolean = false> =
  R extends true
    ? (args: Record<string, unknown>, positional: unknown[], ctx: AuthedCliContext) => Promise<unknown>
    : (args: Record<string, unknown>, positional: unknown[], ctx: CliContext) => Promise<unknown>;
```

Setting `export const requiresAuth = true` buys a non-null `ctx.session` in the handler body without casts. Consumers who want the ergonomics will type their registrar as `CommandRegistrar<true>`.

Runtime enforcement is independent of the types: the `preAction` hook / dispatcher guard resolves the session regardless of whether the consumer annotated the type. Types are a developer-ergonomics layer on top of the runtime guarantee.

## Out of scope

- Auto-persisting on every `ctx.session = …` assignment via setter/Proxy — explicit `saveSession()` keeps control in the consumer's hands and plays nicely with TS.
- Custom resolvers — pure template functions, no session context needed.

## Migration / versioning

Non-breaking minor release:
- New exports: `AuthedCliContext`, `CommandRegistrar<R>`, `DispatcherHandler<R>`, `ctx.resolveSession()`, `ctx.saveSession()`, `.apijack/settings.json` with `customCommands.defaults.requiresAuth`.
- Existing consumers that do not set the flag keep the current behavior (`requiresAuth: false`) and the current nullable `session` type. No changes required to upgrade.
- Docs: add a "Custom commands and auth" section covering the flag, the settings default, the persistence helper, and a before/after example.

## Acceptance criteria

- Setting `export const requiresAuth = true` on a custom command causes core to resolve the session before the action runs.
- Setting `export const requiresAuth = true` on a dispatcher causes core to resolve the session before the handler runs.
- Setting `customCommands.defaults.requiresAuth: true` in `.apijack/settings.json` applies to all custom commands and dispatchers that don't set their own flag.
- Module-level flag overrides settings default.
- `ctx.saveSession()` writes the current `ctx.session` to `~/.<cliName>/session.json` via the wired `SessionManager` instance, with no direct import required by the consumer.
- `ctx.resolveSession()` is a public no-arg alias for the existing private `resolveSession` / `client.ensureReady` path.
- When `requiresAuth` is true, the `AuthedCliContext` type narrowing propagates through `CommandRegistrar<true>` and `DispatcherHandler<true>`.
- Existing consumers (no flag set, no settings key) observe no behavior change.

garretpremo:
Closes #37

## Summary

- Adds `export const requiresAuth = true` for custom commands and dispatchers in `.apijack/commands/*.ts` and `.apijack/dispatchers/*.ts` — the framework resolves `ctx.session` via a `preAction` hook (commands) or handler wrap (dispatchers) before user code runs.
- Adds project-wide default via `<projectRoot>/.apijack/settings.json` under `customCommands.defaults.requiresAuth`. Module-level flag overrides it.
- Adds `ctx.resolveSession()` (public alias for the lazy resolver) and `ctx.saveSession()`. `saveSession()` with `ctx.session === null` calls `SessionManager.invalidate()`.
- Adds `AuthedCliContext` and generic `CommandRegistrar<R>` / `DispatcherHandler<R>` so `R extends true` narrows `ctx.session` to non-null.

## Design decisions (from triage)

- Settings file is project-scoped at `<projectRoot>/.apijack/settings.json`.
- Preview modes (`--dry-run`, `-o curl`, `-o routine-step`, `--help`) skip the preAction hook. `-o curl-with-creds` still resolves.
- `ctx.saveSession()` with null `ctx.session` invalidates the session file.

## Breaking change to note

`loadProjectDispatchers` return type changed from `Map<string, DispatcherHandler>` to `LoadedDispatcher[]` to carry `requiresAuth` metadata. Mirrors the existing `loadProjectCommands` shape. Only internal caller is `bin/apijack.ts`, updated here.

## Follow-up

Wiki entry for Project Mode needs updating. Wiki is gitignored in this repo; draft text is staged locally in `wiki/Project-Mode.md` for reference.

## Test plan

- [x] `bun test` — 731 pass / 0 fail (17 new tests covering the flag, settings resolution, preview-mode skip, `saveSession` persist/invalidate, name-mismatch subcommand).
- [x] `bun run lint` — 0 errors.
- [x] CI E2E green.

### DIFF
--- a/bin/apijack.ts
+++ b/bin/apijack.ts
@@ -10,6 +10,7 @@ import { BearerTokenStrategy } from '../src/auth/bearer';
 import { ApiKeyStrategy } from '../src/auth/api-key';
 import { findProjectConfig, loadProjectConfig, resolveConfigDir } from '../src/project';
 import { loadProjectAuth, loadProjectCommands, loadProjectDispatchers, loadProjectResolvers } from '../src/project-loader';
+import { loadProjectSettings } from '../src/settings';
 import { checkForUpdate } from '../src/updater';
 import { getActiveEnvConfig } from '../src/config';
 import pkg from '../package.json';
@@ -131,7 +132,12 @@ let sessionAuth: SessionAuthConfig | undefined;
     }
 }
 
-// 8. Create CLI
+// 8. Load project settings (framework-level flags, separate from per-env config)
+const projectSettings = projectRoot
+    ? loadProjectSettings(join(projectRoot, '.apijack'))
+    : {};
+
+// 9. Create CLI
 const cli = createCli({
     name: CLI_NAME,
     description: 'Jack into any OpenAPI spec and rip a full-featured CLI',
@@ -142,20 +148,21 @@ const cli = createCli({
     generatedDir,
     allowedCidrs: projectConfig?.allowedCidrs,
     configPath: join(configDir, 'config.json'),
+    customCommandDefaults: projectSettings.customCommands?.defaults,
 });
 
-// 9. Register project-level extensions
+// 10. Register project-level extensions
 if (projectRoot) {
     const commands = await loadProjectCommands(join(projectRoot, '.apijack'));
 
     for (const cmd of commands) {
-        cli.command(cmd.name, cmd.registrar);
+        cli.command(cmd.name, cmd.registrar, { requiresAuth: cmd.requiresAuth });
     }
 
     const dispatchers = await loadProjectDispatchers(join(projectRoot, '.apijack'));
 
-    for (const [name, handler] of dispatchers) {
-        cli.dispatcher(name, handler);
+    for (const disp of dispatchers) {
+        cli.dispatcher(disp.name, disp.handler, { requiresAuth: disp.requiresAuth });
     }
 
     const resolvers = await loadProjectResolvers(join(projectRoot, '.apijack'));
@@ -165,5 +172,5 @@ if (projectRoot) {
     }
 }
 
-// 10. Run
+// 11. Run
 await cli.run();
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -27,9 +27,17 @@ import { resolveRequestHeaders } from './auth/resolve-headers';
 import { deepMergeSessionAuth } from './auth/config-merge';
 import { loadPreRequestHook } from './pre-request';
 
+export interface CommandOptions {
+    requiresAuth?: boolean;
+}
+
+export interface DispatcherOptions {
+    requiresAuth?: boolean;
+}
+
 export interface Cli {
-    command(name: string, registrar: CommandRegistrar): void;
-    dispatcher(name: string, handler: DispatcherHandler): void;
+    command(name: string, registrar: CommandRegistrar, options?: CommandOptions): void;
+    dispatcher(name: string, handler: DispatcherHandler, options?: DispatcherOptions): void;
     resolver(name: string, handler: CustomResolver): void;
     run(): Promise<void>;
 }
@@ -102,8 +110,8 @@ function showCustomHelp(
 }
 
 export function createCli(options: CliOptions): Cli {
-    const consumerCommands: { name: string; registrar: CommandRegistrar }[] = [];
-    const consumerDispatchers = new Map<string, DispatcherHandler>();
+    const consumerCommands: { name: string; registrar: CommandRegistrar; requiresAuth?: boolean }[] = [];
+    const consumerDispatchers = new Map<string, { handler: DispatcherHandler; requiresAuth?: boolean }>();
     const consumerResolvers = new Map<string, CustomResolver>();
     const cliName = options.name;
     const configOpts = options.configPath ? { configPath: options.configPath } : undefined;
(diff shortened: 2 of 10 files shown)


garretpremo (commented):
Reviewed against #37 acceptance criteria. All 8 criteria are implemented and CI is green (lint, unit-tests, edge cases, OAS 3.1, Petstore across ubuntu + windows). Tests are comprehensive — the alias-name edge case is nicely covered by walking up to the top-level Command and matching on instance identity rather than name.

Three non-blocking observations:

1. **Dispatcher path has no preview-mode skip.** `cli-builder.ts` guards the commands `preAction` hook with `!skipAuthResolution` (skips `--dry-run`, `-o curl`, `-o routine-step`, `--help`). The dispatcher wrapper around line 488 unconditionally calls `resolveSession()` when `requiresAuth` is true. Dispatchers are driven through routines, so in practice this may be fine, but the behavior is asymmetric with commands and worth a comment or a follow-up if a routine's dry-run path ever reaches a requiresAuth dispatcher.

2. **`-o routine-step skips requiresAuth preAction hook` test (`tests/custom-command-auth.test.ts`) is weak.** It asserts `actionRan === false`, which is true regardless of whether the hook fires — `-o routine-step` exits before the action runs. A stronger assertion would spy on `resolveSession` / session state to prove the hook was skipped.

3. **Unrelated cast:** `client: ctx.client as Record<string, unknown> | undefined` in the `buildDispatcher` call looks outside the scope of this PR. If it was needed for compilation, a short comment explaining why would help future readers.

No changes requested — these are follow-up worthy, not merge blockers.

PR #41, opened 2026-04-23
feat(resolver): balanced-delimiter matcher unlocks nested and multi-line function args

garretpremo:
## Summary

Replaces the regex-based `$_foo(...)` detector in `src/routine/resolver.ts` with a hand-written balanced-delimiter matcher. Lifts the single-level, single-line restrictions of the current resolver.

**Foundation for the v1.9.0 plugin protocol** — pre-built plugins (faker, dayjs, etc.) need multi-line JSON5 object args and nested `$_foo(..., $_bar(...))` calls, both blocked by the old regex.

## What changes

- **New matcher module** (`src/routine/resolver-matcher.ts`) — pure, hand-written, respects `()`, `{}`, `[]`, and single/double-quoted strings with `\`-escapes. Exports `findFunctionCalls()` and `isExactFunctionCall()`.
- **`resolver.ts` rewired** — `resolveValue` and `resolveString` use the matcher instead of `PARAM_FUNC_PATTERN` / `EXACT_PARAM_FUNC_PATTERN`. `NOARG_FUNC_PATTERN` and `EXACT_NOARG_FUNC_PATTERN` retained for `$_uuid`-style zero-arg functions. Resolution order preserved: parameterized → no-arg → refs.
- **Backward-compatible passthrough** — malformed function-call syntax (unclosed parens, etc.) is silently passed through, matching the old regex's behavior exactly. Verified in tests.

## What this unlocks

```yaml
# Multi-line JSON5 args (previously broken — regex stopped at first `)`)
--user: |
  $_faker(person.fullName, {
    firstName: "Alex",
    sex: "female"
  })

# Nested function calls (previously broken — regex didn't balance parens)
--email: '$_faker(internet.email, {firstName: "$_faker(person.firstName)"})'

# Close-parens inside quoted string args
--note: '$_mkComment("finished (pass 1)")'
```

## Test plan

- [x] New matcher module: 20 unit tests (`tests/routine/resolver-matcher.test.ts`) — single/nested calls, multi-line, escaped quotes, mismatched delimiters, unclosed strings/parens
- [x] Resolver integration: 5 new tests in `tests/routine/resolver.test.ts` — nested calls via `helpers.resolve`, multi-line args, quoted-string close-paren, braces/brackets, graceful malformed-input passthrough
- [x] Full suite: **756 pass / 0 fail** (was 751 on dev; +5 net)
- [x] Old regex behavior preserved byte-for-byte on malformed inputs (verified empirically in review)

## Spec reference

`docs/superpowers/specs/2026-04-23-prebuilt-plugins-design.md` §5.6 (resolver core upgrade).

### DIFF
--- a/src/routine/resolver-matcher.ts
+++ b/src/routine/resolver-matcher.ts
@@ -75,7 +75,7 @@ function findMatchingParen(input: string, openIdx: number): number {
         j++;
     }
 
-    throw new Error(`unclosed parenthesis starting near index ${openIdx - 1}`);
+    throw new Error(`unclosed parenthesis starting at index ${openIdx}`);
 }
 
 export function findFunctionCalls(input: string): CallMatch[] {
@@ -102,7 +102,17 @@ export function findFunctionCalls(input: string): CallMatch[] {
             continue;
         }
 
-        const closeIdx = findMatchingParen(input, nameEnd);
+        let closeIdx: number;
+
+        try {
+            closeIdx = findMatchingParen(input, nameEnd);
+        } catch {
+            // Malformed: unclosed paren or unclosed string literal inside args.
+            // Skip past the opening `(` and keep scanning for further valid calls.
+            i = nameEnd + 1;
+            continue;
+        }
+
         const name = input.slice(i + 1, nameEnd);
         const argsStr = input.slice(nameEnd + 1, closeIdx);
         results.push({ name, argsStr, start: i, end: closeIdx + 1 });
@@ -113,17 +123,15 @@ export function findFunctionCalls(input: string): CallMatch[] {
 }
 
 export function isExactFunctionCall(input: string): { name: string; argsStr: string } | null {
-    const trimmed = input;
+    if (input[0] !== '$' || input[1] !== '_') return null;
 
-    if (trimmed[0] !== '$' || trimmed[1] !== '_') return null;
-
-    const matches = findFunctionCalls(trimmed);
+    const matches = findFunctionCalls(input);
 
     if (matches.length !== 1) return null;
 
     const m = matches[0]!;
 
-    if (m.start !== 0 || m.end !== trimmed.length) return null;
+    if (m.start !== 0 || m.end !== input.length) return null;
 
     return { name: m.name, argsStr: m.argsStr };
 }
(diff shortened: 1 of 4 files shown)


garretpremo (commented):
LGTM — spec §6 acceptance criteria met, matcher correctly replaces both regex patterns, 756/756 tests pass on ubuntu+windows, lint clean on touched files. Safe to merge.

<details>

- **"Byte-for-byte" claim in the PR body is slightly off.** When a string contains *both* a valid parameterized call and a malformed one, the behaviors diverge because `findFunctionCalls` is invoked on the whole string — any throw skips the splice pass for the whole string, then the no-arg pass picks up the valid call's identifier *without* its args:

  ```
  Input:  $_valid(a) $_foo(unclosed
  dev:    V[a] (unclosed           ← PARAM pass resolved $_valid(a)
  PR:     V[undefined](a) (unclosed ← matcher threw, no-arg pass called _valid with undefined args
  ```

  Not a real-world regression (malformed routines aren't supported either way) but the changelog/PR body overstates compatibility. If you want true per-call resilience, have the matcher skip individual malformed spans instead of throwing from the whole scan.

</details>

<details>
<summary>Nitpicks</summary>

- `resolver-matcher.ts:117` — `const trimmed = input;` is a misleading name since nothing is trimmed. Just use `input` directly.
- `resolver-matcher.ts:66` — the error message `"unclosed parenthesis starting near index ${openIdx - 1}"` points at the identifier's last char rather than the `(`. Errors are caught by callers so it's only ever seen during debugging, but `openIdx` itself would be more accurate.

</details>

Reviewed by claude 🤖

garretpremo:
Addressed the review in `9a055e7`:

- **Byte-for-byte parity restored.** Refactored `findFunctionCalls` to skip past malformed `$_name(` spans internally instead of throwing on the whole scan. Mixed valid+malformed inputs now produce the same output as the old regex — locked in by a new test `mixed valid + malformed: valid call resolves, malformed span passes through (old-regex parity)`.
- **Resolver try/catch wrappers dropped.** No longer needed since the matcher doesn't throw on malformed input.
- **Nit 1 fixed:** removed dead `const trimmed = input;` in `isExactFunctionCall`.
- **Nit 2 fixed:** unclosed-paren error message now points at `openIdx` (the `(` itself) instead of `openIdx - 1`.

Two prior matcher throw-tests rewritten to assert skip-past behavior; 2 new matcher tests assert skip still matches preceding/subsequent valid calls. Full suite: 759/759.

PR #42, opened 2026-04-23
feat(plugin): ApijackPlugin contract, cli.use(), and startup validation

garretpremo:
## Summary

Second wave of the v1.9.0 plugin protocol. Adds the typed `ApijackPlugin` contract, `cli.use()` registration, and startup validation covering namespace rules, resolver-name collisions, and peer-dep semver compatibility.

**Builds on #41** (balanced-delimiter matcher). No user-facing commands yet — diagnostic commands (`plugins list` / `plugins check`) and routine-scoped plugin configuration come in Wave 3.

## What changes

**New `ApijackPlugin` interface** (`src/types.ts`):

```ts
export interface ApijackPlugin {
    name: string;                               // lowercase identifier, required prefix
    version?: string;
    resolvers?: Record<string, CustomResolver>;  // stateless, merged into routine context
    createRoutineResolvers?: (opts) => Record<...>;  // per-routine factory (wired in later wave)
    __package?: { name: string };               // npm package name, for peer-version lookup
}
```

**`cli.use(plugin)` registration** (`src/cli-builder.ts`) — explicit, typed, returns `void`. Plugins accumulate in an internal `PluginRegistry`.

**Startup validation** — runs as step 0 of `cli.run()`, before any command setup:

1. **Plugin name** must match `/^[a-z][a-z0-9_]*$/`. Rejected at registration.
2. **Namespace enforcement** — plugin named `faker` can register `_faker` and `_faker_*` only; `_fakerish` or `_other` throw `PluginNamespaceError`.
3. **Collision detection** — plugin resolver keys cannot collide with core built-ins (`_uuid`, `_random_int`, `_find`, etc.) or project resolvers from `.apijack/resolvers/`. Throws `PluginCollisionError`.
4. **Peer-version check** — if `plugin.__package` is set, the loader walks `node_modules` looking for the plugin's `package.json` and reads `peerDependencies[\"@apijack/core\"]`. Mismatch → `PluginPeerMismatchError`. Missing `__package` → stderr warning, check skipped.

**Single source of truth for built-in names** — `BUILTIN_RESOLVER_NAMES` now exported from `src/routine/resolver.ts`; `src/plugin/registry.ts` and `src/project-loader.ts` import it. No more manual drift across files.

**Stateless plugin resolvers merged into routine context** — plugins with a `resolvers` map have those functions made available to every routine via the existing `customResolvers` pipeline.

## Why this shape

- **Explicit over auto-discovery** — `cli.use(plugin)` makes the load set obvious in code review; no package.json scraping.
- **Startup validation (fail fast)** — collisions, malformed names, and peer mismatches are caught once at program start, not intermittently during dispatch.
- **Namespaced resolvers** — plugins can only register keys under their own name (`_<name>`, `_<name>_*`), so plugin-vs-plugin collision is structurally impossible.
- **Major-version coupling** — `peerDependencies: { \"@apijack/core\": \"^1.0.0\" }` (semver range) enforced at startup. Every plugin major tracks core major in lockstep.

## Test plan

- [x] 3 error classes with structured fields (namespace / collision / peer-mismatch)
- [x] Registry: registration, dup-name, name-format, namespace violations, collision detection (core-builtin + project-resolver), factory invoked at most once per `validateAll`
- [x] Peer-version: ranges (`^`, `~`, `>=`), invalid ranges, missing declarations, scoped npm names, invalid JSON tolerance, parent-directory walk
- [x] `cli.use()`: accepts plugins, rejects duplicate names
- [x] `cli.run()` validation: throws `PluginNamespaceError` / `PluginCollisionError`, warns on missing `__package`
- [x] Full suite: **797 pass / 0 fail** (was 759 on dev; +38 net)

## Spec reference

`docs/superpowers/specs/2026-04-23-prebuilt-plugins-design.md` §§5.1–5.5, §5.8.


garretpremo (commented):
**Review outcome: changes requested** (couldn't formally request changes on own PR — label reflects outcome.)

CI is red on Windows — one peer-version test fails due to a path-separator assumption in the test. All other test-plan items look good; the design matches the spec and the code is tight. Blocking on the CI failure.

**Blocking:**
- `unit-tests (windows-latest)` fails on `tests/plugin/peer-version.test.ts` → `loadPluginPeerInfo > reads peerDependencies for a scoped package name`. The assertion `expect(info.packagePath).toContain('@scope/plugin')` (line 127) uses a forward-slash literal, but Windows `path.join` produces `@scope\plugin`. Make the assertion separator-agnostic — e.g. `expect(info.packagePath).toMatch(/[\\\/]@scope[\\\/]plugin[\\\/]/)` or compare against `join('@scope', 'plugin')`. (The `'walks up parent directories'` test happens to only assert on `declaredRange`, so it doesn't trip the same issue.)

<details>

- `cli.run()` does the peer-version `node_modules` walk starting from `process.cwd()` (cli-builder.ts:162). For a globally-installed CLI with plugins installed alongside, the plugin package lives near the CLI binary, not the user's cwd — the walk won't find it and the check silently no-ops (only the missing-`__package` warning fires, not a missing-package-on-disk one). Consider starting from `__dirname` / the CLI's module dir, or walking both. Probably acceptable for Wave 2, but worth noting before Wave 3 ships user-facing `plugins check`.
- `PluginRegistry.register()` throws plain `Error` for invalid name / duplicate name (registry.ts:17, :23) while namespace/collision/peer use typed classes. Consider a `PluginRegistrationError` for symmetry so consumers can catch all plugin-setup failures uniformly.
- `collectPluginInfo` silently swallows throws from `createRoutineResolvers({})` (registry.ts:52–60). Pragmatic, but a stderr warning would aid debugging — otherwise a plugin whose factory throws on `{}` quietly skips namespace/collision checks on its dynamic keys.
- `__package` is typed as `{ name: string }` only (types.ts:70). If future waves want the plugin's own `version` (e.g., for `plugins list`), widening the type now avoids a breaking change later.

</details>

<details>
<summary>Nitpicks</summary>

- peer-version.ts:37 doc says \"or has no declared peer range\" but when the package exists without a peer dep, `packagePath` is still truthy — only `declaredRange` is `undefined`. Minor wording inconsistency.
- registry.ts:58 comment \"Plugin rejected the dry call; skip factory-output checks.\" is narrating the obvious try/catch; safe to drop.

</details>

Reviewed by claude 🤖

garretpremo (commented):
**Review outcome: changes requested** (couldn't formally request changes on own PR — posting as a comment. Label reflects outcome.)

Prior-round items are addressed — Windows CI is green, the peer-lookup now walks from both `process.cwd()` and `import.meta.dir`, `PluginRegistrationError` rounds out the error hierarchy, and the `collectPluginInfo` catch now surfaces a stderr warning. One new blocking issue surfaced on this read: the peer-version check compares against the consumer CLI's version, not `@apijack/core`'s version.

**Blocking:**
- `cli-builder.ts:165,172` passes `options.version` as `installedVersion` to `checkPeerRange` / `PluginPeerMismatchError`, but the plugin's peer range is declared on `@apijack/core`. For the apijack binary itself these happen to be equal because `bin/apijack.ts` does `import pkg from '../package.json'; version: pkg.version` — that `pkg` is `@apijack/core`'s own — so every existing test passes. For external consumers (`createCli({ version: "2.3.0" })` where `2.3.0` is the consumer's CLI version, not the core version) the comparison is wrong: a plugin declaring `peerDependencies["@apijack/core"]: "^1.0.0"` against installed core 1.9.0 would be compared against `2.3.0` and throw a spurious `PluginPeerMismatchError`. The error message `"installed core is ${installedVersion}"` (errors.ts:35) encodes the intended semantics clearly. Fix: read core's version at load time in `cli-builder.ts` (`import pkg from '../package.json' with { type: 'json' }` — that file resolves to `@apijack/core`'s own package.json since cli-builder.ts lives inside the package) and pass `pkg.version` as `installedVersion` instead of `options.version`. Keep `options.version` only for the Commander `--version` flag.

<details>

- No integration test exercises the peer-version path end-to-end — a plugin with a real `__package`, a fake `node_modules` tree under `process.cwd()` or `import.meta.dir`, and `cli.run()` asserting mismatch vs. satisfy. Unit tests cover `checkPeerRange` and `loadPluginPeerInfo` individually, and the cli.run() tests exercise namespace + collision + missing-`__package`, but the wiring bug above lives in the gap between those layers. Worth adding before Wave 3 ships `plugins check`, which will lean on the same path.
- `PluginCollisionError` sourceB is hardcoded as `".apijack/resolvers/ project resolver"` (registry.ts:103–105), but `validateCollisions` receives `consumerResolvers` which also includes programmatically-registered resolvers from `cli.resolver(name, fn)`. A collision against a consumer-code resolver would incorrectly blame the `.apijack/resolvers/` directory.
- `peer-version.ts` silently treats invalid JSON in package.json as "no declared range" (catch at lines 40–42 of the new file). A stderr warning — same pattern as the new `collectPluginInfo` warning — would aid debugging when a plugin's package.json is malformed.
- `__package` is still typed as `{ name: string }` only (types.ts:71), carried from the prior round's nit. Wave 3's `plugins list` is going to want `version` too — widening to `{ name: string; version?: string }` in this wave avoids a later breaking change.

</details>

<details>
<summary>Nitpicks</summary>

- The "warns to stderr when plugin has no __package" test (cli-builder.test.ts:568–607) does manual monkeypatching of `process.stderr.write` / `stdout.write` / `console.log` / `process.exit` and uses `.catch(() => {})` to swallow unrelated downstream errors. `spyOn(process.stderr, 'write')` scoped to just the validation pass would be cleaner and more focused on what's actually under test.
- `collectPluginInfo` warning says "skipping dynamic resolver validation" (registry.ts:56) — "dynamic" isn't user-facing terminology. "skipping createRoutineResolvers-output validation" reads more directly.

</details>

Reviewed by claude 🤖

garretpremo (commented):
All prior-round blocking issues are resolved and CI is fully green on both ubuntu and windows. The peer-version check now correctly reads `@apijack/core`'s own version from `import.meta.dir/../package.json` (cli-builder.ts:35-37, used at 170 and 177), not the consumer CLI's version. The new `cli.run()` integration tests in `cli-builder.test.ts` exercise both the success path (consumer version `99.0.0`, peer range pinned to core's major → no mismatch) and the failure path (`^999.0.0` peer range → throws `PluginPeerMismatchError`), closing the wiring gap flagged in the last round.

<details>

- The peer-version success-path integration test asserts with `expect(String(thrown)).not.toContain('PluginPeerMismatch')` (cli-builder.test.ts:522). That passes trivially if `thrown` ends up `null` from the success path, which is the intended behavior, but it also passes for any unrelated error that doesn't stringify with that substring. Tightening to `expect(thrown).toBeNull()` or `expect(thrown).not.toBeInstanceOf(PluginPeerMismatchError)` would make the intent read more directly.
- `PluginRegistry.validateCollisions` still blames "consumer resolver" generically (registry.ts:104) — which is correct now after the round-2 fix. Worth noting for Wave 3: once plugins can collide with each other's keys too (if the namespace ever loosens), this error would need a plugin-vs-plugin source label. Not actionable here; just a flag.

</details>

<details>
<summary>Nitpicks</summary>

- The \"warns to stderr when plugin has no __package\" test (cli-builder.test.ts:468-509) still uses manual monkeypatching of `process.stderr.write` / `stdout.write` / `console.log` / `process.exit` instead of scoped `spyOn`. Carried from last round — nit, not blocking.
- `registry.ts:64` still reads `skipping dynamic resolver validation`; \"dynamic\" isn't the user-facing term. Swapping to `skipping createRoutineResolvers-output validation` reads more directly. Carried from last round.

</details>

Reviewed by claude 🤖

PR #43, opened 2026-04-23
feat(routine): plugins: YAML field + per-routine scoping + sub-routine inheritance

garretpremo:
## Summary

Third wave of the v1.9.0 plugin protocol. Adds the routine-level `plugins:` YAML field, a `buildRoutineResolvers` helper that merges global + plugin resolvers per routine, and executor wiring that gives each routine invocation a fresh closure from each plugin's `createRoutineResolvers` factory. Sub-routines inherit parent scope by default and override only when they declare their own `plugins:` block.

**Builds on #42.** Completes the plumbing side of the plugin protocol — routines can now configure plugins and plugins can observe per-routine state. User-facing commands (`plugins list` / `plugins check`) come in Wave 4.

### New routine field

```yaml
name: seeded-user-gen
plugins:
  faker:
    seed: 42
    locale: en
steps:
  - name: make-user
    command: users create
    args:
      --name: "$_faker(person.fullName)"
```

- `RoutineDefinition.plugins?: Record<string, unknown>` added.
- Loader whitelist updated to pass the field through unchanged.
- Shape validation is deferred to each plugin (the blob is opaque at the routine layer per spec §7).

### New `buildRoutineResolvers` helper (`src/routine/plugin-resolvers.ts`)

Pure, testable merge:

1. Start from global resolvers (project `.apijack/resolvers/` + programmatic `cli.resolver()`).
2. Add each plugin's stateless `plugin.resolvers`.
3. Call each plugin's `createRoutineResolvers(routine.plugins[name] ?? {})` and merge the returned map — fresh closure per routine.

Emits stderr warnings (non-fatal) on:
- Unknown plugin key in `routine.plugins:` (warning names the routine and the offending key).
- `createRoutineResolvers` throw (warning names the routine, plugin, and error message).

### Executor wiring

- `ExecutorOptions.pluginRegistry?: PluginRegistry` — threaded from `cli.use()` call site through `buildDispatcher` → `registerRoutineCommand` → `routineRunAction` / `routineTestAction` → `executeRoutine` on every code path.
- `RoutineExecutor.execute` calls `buildRoutineResolvers` at routine start to build `ctx.customResolvers`.

### Sub-routine scoping

In `src/routine/dispatcher.ts`, the `routine run` branch gates registry forwarding on whether the sub-routine has its own non-empty `plugins:` block:

- **Sub has `plugins:`** → pass registry → factories re-invoked with sub's opts → fresh closures that shadow parent's for the subtree.
- **Sub has no `plugins:`** → don't pass registry → `buildRoutineResolvers` returns parent's resolver map unchanged → sub inherits parent's closures.

Edge cases handled: `plugins: null`, `plugins: {}`, `plugins: { faker: null }` all treated as "no configuration" (inherit path).

### Minor polish (folded in from review)

- `createRoutineResolvers` throw warning now includes routine name for debuggability.

## Why this shape

- **`buildRoutineResolvers` is a pure helper** — all the merge semantics live in one testable function instead of scattered across the executor and dispatcher. The dispatcher's only job is deciding *whether* to pass the registry; everything else is uniform.
- **Inherit-by-default for sub-routines** matches spec §8 and preserves a parent's seeded faker state across sub-routine boundaries unless explicitly overridden.
- **Backward compatible.** `pluginRegistry` is optional everywhere; existing `executeRoutine(routine, overrides, dispatch)` calls without the options keep working (covered by a dedicated test).

## Test plan

- [x] `buildRoutineResolvers`: stateless merge, factory invocation, opts pass-through, default `{}`, per-call closures, unknown-key warning, global preservation, undefined registry, combined stateless + dynamic plugins, factory-throw warning with routine name
- [x] Sub-routine scoping: inherit without `plugins:`, override with `plugins:` (asserts actual closure replacement via `tag`-based output, not just map key presence)
- [x] Executor integration: factory invoked once per `executeRoutine` call; two routine runs produce independent closures; backward compat without a registry
- [x] Loader: `plugins:` field parsed + passed through; absent field defaults to `undefined`
- [x] Full suite: **818 pass / 0 fail** (was 812 at branch start; +6 net)

## Spec reference

`docs/superpowers/specs/2026-04-23-prebuilt-plugins-design.md` §7 (routine `plugins:` field) and §8 (per-routine resolver scoping).

### DIFF
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -557,6 +557,7 @@ export function createCli(options: CliOptions): Cli {
                     client: ctx.client as Record<string, unknown> | undefined,
                     consumerHandlers: dispatcherHandlers,
                     customResolvers,
+                    pluginRegistry,
                     preDispatch: options.preDispatch,
                     ctx,
                     routinesDir,
@@ -567,7 +568,7 @@ export function createCli(options: CliOptions): Cli {
             }
 
             // 12. Register routine commands
-            registerRoutineCommand(program, cliName, routinesDir, dispatch, options.builtinRoutinesDir, customResolvers);
+            registerRoutineCommand(program, cliName, routinesDir, dispatch, options.builtinRoutinesDir, customResolvers, pluginRegistry);
 
             // 13. Handle -o routine-step
             if (isRoutineStep) {
--- a/src/commands/routine/register.ts
+++ b/src/commands/routine/register.ts
@@ -2,6 +2,7 @@ import { Command } from 'commander';
 import { existsSync, mkdirSync, cpSync, readdirSync, readFileSync } from 'fs';
 import { resolve } from 'path';
 import type { CommandDispatcher, CustomResolver } from '../../types';
+import type { PluginRegistry } from '../../plugin/registry';
 import { SessionManager } from '../../session';
 import { loadRoutineFile, loadSpecFile, listRoutines, validateRoutine, formatRoutineTree, formatRoutineList } from '../../routine/loader';
 import { executeRoutine } from '../../routine/executor';
@@ -44,6 +45,7 @@ export function registerRoutineCommand(
     dispatch: CommandDispatcher | undefined,
     builtinRoutinesDir?: string,
     customResolvers?: Map<string, CustomResolver>,
+    pluginRegistry?: PluginRegistry,
 ): void {
     const builtinsMap = builtinRoutinesDir
         ? loadBuiltinRoutines(builtinRoutinesDir)
@@ -116,6 +118,7 @@ export function registerRoutineCommand(
                     overrides,
                     dryRun: opts.dryRun,
                     customResolvers,
+                    pluginRegistry,
                     invalidateSession: () => sessionMgr.invalidate(),
                     onStep: (step, i, total) => {
                         console.log(`\x1b[36m[${i + 1}/${total}]\x1b[0m ${step.name}`);
@@ -188,6 +191,7 @@ export function registerRoutineCommand(
                     dispatch,
                     overrides,
                     customResolvers,
+                    pluginRegistry,
                     routineName: name,
                     onStep: (step, i, total) => {
                         console.log(
--- a/src/commands/routine/run/run.ts
+++ b/src/commands/routine/run/run.ts
@@ -1,15 +1,17 @@
 import type { CommandDispatcher, CustomResolver } from '../../../types';
+import type { PluginRegistry } from '../../../plugin/registry';
 import type { RoutineDefinition, RoutineStep } from '../../../routine/types';
 import type { RoutineResult } from '../../../routine/executor';
 
 export interface RoutineRunDeps {
     loadRoutine: () => RoutineDefinition;
     validateRoutine: (def: RoutineDefinition) => string[];
-    executeRoutine: (def: RoutineDefinition, overrides: Record<string, unknown>, dispatch: CommandDispatcher, opts: { dryRun?: boolean; customResolvers?: Map<string, CustomResolver>; onStep?: RoutineRunDeps['onStep']; onIteration?: RoutineRunDeps['onIteration'] }) => Promise<RoutineResult>;
+    executeRoutine: (def: RoutineDefinition, overrides: Record<string, unknown>, dispatch: CommandDispatcher, opts: { dryRun?: boolean; customResolvers?: Map<string, CustomResolver>; pluginRegistry?: PluginRegistry; onStep?: RoutineRunDeps['onStep']; onIteration?: RoutineRunDeps['onIteration'] }) => Promise<RoutineResult>;
     dispatch: CommandDispatcher;
     overrides: Record<string, unknown>;
     dryRun?: boolean;
     customResolvers?: Map<string, CustomResolver>;
+    pluginRegistry?: PluginRegistry;
     invalidateSession: () => void;
     onStep?: (step: RoutineStep, i: number, total: number) => void;
     onIteration?: (step: RoutineStep, current: number, total: number, stepIndex: number, stepTotal: number) => void;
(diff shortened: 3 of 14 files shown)


garretpremo (commented):
Sub-routine scoping doesn't match spec §8's inherit-parent requirement: the dispatcher passes the CLI-level `config.customResolvers` into sub-routines rather than the parent routine's `ctx.customResolvers`, so factory-produced closures registered by the parent do not flow into a sub-routine that lacks its own `plugins:` block.

**Blocking:**
- `src/routine/dispatcher.ts:190-193` forwards `customResolvers: config.customResolvers` (built once at CLI start from consumer resolvers + stateless `plugin.resolvers`) to the sub's executor. It has no access to the parent routine's `ctx.customResolvers`, which is where `createRoutineResolvers(opts)` output lives. As a result, the canonical spec scenario — parent declares `plugins: { faker: { seed: 42 } }`, sub-routine calls `$_faker(...)` without its own `plugins:` block — will fail: `_faker` is produced only via `createRoutineResolvers` (faker plugin defines no stateless `resolvers`), so the sub's resolver map has no `_faker` entry at all. Spec §8 is explicit: *"If the sub-routine has no `plugins:` block, inherit the parent's `customResolvers` directly."* The PR body promises this (`sub inherits parent's closures`) but the wiring doesn't deliver it.
- The unit tests `sub without plugins: inherits parent map` and `sub with plugins: triggers fresh factory call` in `tests/plugin/routine-scoping.test.ts` pass `parentMap` as the `globalResolvers` argument to `buildRoutineResolvers`, simulating inheritance that the real dispatcher never performs. `buildRoutineResolvers` in isolation is correct; the integration isn't. An end-to-end test that runs a parent routine through the real dispatcher and asserts a sub-routine sees the parent's seeded plugin closure would fail today and should be added alongside the fix.

**Suggested fix direction:** thread the parent's ctx (or at least its `customResolvers`) through the sub-routine invocation path — e.g., have `RoutineExecutor` pass `ctx.customResolvers` to `dispatch` when issuing `routine run`, and let the dispatcher prefer that over `config.customResolvers` when present. Alternative: expose a small "enter sub-routine" API on the dispatcher that takes the parent's resolver map.

<details>

- The sub-routine-scoping comment in `src/routine/dispatcher.ts:181-187` asserts that `buildRoutineResolvers` "returns the parent's resolver map unchanged" when the registry is omitted — under the current wiring it returns `config.customResolvers` (the CLI-global map), not the parent's per-routine map. Please update the comment to match whichever semantics the fix lands on.
- Spec §10.4 suggests `tests/routine/plugins-config.test.ts`; the tests here live at `tests/plugin/routine-scoping.test.ts`. Not blocking, but worth noting if you want to keep file layout aligned with the spec doc.
- PR body claims "818 pass / 0 fail" but `bun test` locally reports 817 pass. Minor counting skew — not a correctness concern.

</details>

Reviewed by claude 🤖

garretpremo (commented):
The blocking issue from the previous review (2026-04-23T16:41:15Z on commit `bfe9945`) is unchanged on this same commit — spec §8's "inherit the parent's `customResolvers` directly" path is still not wired end-to-end through the real dispatcher.

**Blocking:**
- `src/routine/dispatcher.ts:190-193` still forwards `customResolvers: config.customResolvers` (the CLI-global map built once at `cli.run()` — consumer resolvers plus stateless `plugin.resolvers` only) into the sub's executor. The parent routine's `ctx.customResolvers`, which is where `createRoutineResolvers(opts)` output lives (e.g. a seeded `_faker`), is not available to this call site. Spec §8 L211: *"If the sub-routine has no `plugins:` block, inherit the parent's `customResolvers` directly."* The canonical scenario — parent declares `plugins: { faker: { seed: 42 } }`, sub-routine calls `$_faker(...)` with no `plugins:` block — still fails: `_faker` is produced only via `createRoutineResolvers`, so the sub's resolver map has no `_faker` entry.
- `tests/plugin/routine-scoping.test.ts` (`sub without plugins: inherits parent map`, `sub with plugins: triggers fresh factory call`) still pass `parentMap` directly as the `globalResolvers` argument to `buildRoutineResolvers`. That simulates inheritance the dispatcher never performs. `buildRoutineResolvers` in isolation remains correct; the integration is untested. Please add an end-to-end test that drives a parent routine through the real dispatcher + executor and asserts a sub-routine sees the parent's seeded plugin closure — that test should fail against the current code and pass after the fix.

**Suggested fix direction (unchanged from prior review):** thread the parent's ctx (or just its `customResolvers` map) through the sub-routine invocation — e.g., extend `CommandDispatcher` with an optional `routineCtx` param so `RoutineExecutor.executeCommand` can pass `ctx` along to `dispatch`, and have the dispatcher prefer `routineCtx.customResolvers` over `config.customResolvers` when present.

<details>

- `src/routine/dispatcher.ts:181-187` comment still reads "returns the parent's resolver map unchanged" — under the current wiring it returns `config.customResolvers` (the CLI-global map), not the parent's per-routine map. Please keep the comment in sync with whichever semantics the fix lands on.
- Spec §10.4 suggests `tests/routine/plugins-config.test.ts`; the tests here live at `tests/plugin/routine-scoping.test.ts`. Not blocking, but worth noting if you want to keep file layout aligned with the spec doc.
- PR body claims "818 pass / 0 fail" but `bun test` locally reports 817 pass. Minor counting skew — not a correctness concern.

</details>

Reviewed by claude 🤖

garretpremo (commented):
Blocking issue from the prior review (on `bfe9945`) is resolved in `502c6d5`. `CommandDispatcher` now takes an optional `routineCtx` 4th arg; `RoutineExecutor.executeCommand` passes the parent's `ctx` to `dispatch`; the dispatcher's `routine run` branch prefers `routineCtx.customResolvers` over `config.customResolvers` and gates registry passthrough on `subHasPlugins`. Spec §8's "inherit the parent's `customResolvers` directly" path is now wired end-to-end, and a new integration test in `tests/routine/dispatcher.test.ts` drives a parent through the real dispatcher + executor and asserts both scenarios — sub without `plugins:` inherits (`['42', '42']`), sub with its own `plugins:` overrides its subtree only (`['42', '999', '42']`). The dispatcher comment now correctly reflects the inherit/override semantics. All CI checks pass.

<details>

- Spec §10.4 suggests `tests/routine/plugins-config.test.ts`; tests are split across `tests/plugin/routine-scoping.test.ts` (unit) and `tests/routine/dispatcher.test.ts` (integration). Not blocking, just noting for future file layout alignment.

</details>

Reviewed by claude 🤖

PR #44, opened 2026-04-23
feat(plugins): <cli> plugins list + plugins check + docs

garretpremo:
## Summary

Final wave of the v1.9.0 plugin protocol. Adds two diagnostic commands — `<cli> plugins list` and `<cli> plugins check` — plus CLAUDE.md/README documentation for the whole plugin system. Together with #41, #42, #43, this completes the core side of the v1.9.0 plugin protocol.

## Naming decision

The existing `<cli> plugin` command (singular, in `src/plugin/register.ts`) is for registering apijack *as* a Claude Code plugin (install/uninstall). The new diagnostics live under **`<cli> plugins`** (plural) to avoid collision. Both surfaces coexist cleanly — Commander treats them as distinct subcommands.

## What changes

**`<cli> plugins list`** — prints a table of registered apijack plugins:

```
NAME    VERSION  STATUS
faker   1.0.0    ok
dayjs   0.9.0    ok
```

- Missing `version` field shown as `-`, never `undefined`.
- Empty registry prints `No plugins registered.`
- Exits 0.

**`<cli> plugins check`** — runs startup validation on demand:

- Exits 0 with `All plugins OK.` when namespace + collision + peer-version rules pass.
- Exits 1 and prints each failure (namespace violation, core-builtin collision, peer mismatch) on its own line.
- Intended for CI — catches issues without having to actually execute a command.

**Docs** — `CLAUDE.md` gets a new "Plugin System" section with installation, routine-level configuration, and diagnostic-command references. `README.md` gets a short Plugins blurb linking to `CLAUDE.md`.

## Test plan

- [x] `plugins list`: table formatting, empty-registry case, plugin without `version` shows `-`
- [x] `plugins check`: exit 0 when OK, exit 1 on namespace violation, exit 1 on core-builtin collision, exit 0 on empty registry
- [x] Full suite: **826 pass / 0 fail** (was 819; +7 net)

## Spec reference

`docs/superpowers/specs/2026-04-23-prebuilt-plugins-design.md` §6 (Diagnostics).

## What's next

Core v1.9.0 is ready to ship once this merges. Then: `@normalled/apijack-plugin-faker` v1.0.0 in a new repo, tracked by `docs/superpowers/plans/2026-04-23-faker-plugin.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,6 +78,7 @@ steps:
 - **forEach loops**: Iterate over arrays from previous step output
 - **Assertions**: `assert:` validates response fields with `equals`, `contains`, `matches`
 - **Sub-routines**: `routine:` field runs another routine inline
+- **Plugins**: Top-level `plugins:` block passes per-routine configuration to registered apijack plugins (e.g., `plugins: { faker: { seed: 42 } }`). See `<cli> plugins list` for what's installed.
 
 ### Built-in resolver functions
 
@@ -99,6 +100,47 @@ Usable anywhere a routine value is resolved (args, conditions, variables):
 <cli> routine validate <name>   # Validate YAML structure
 <cli> routine test <name>       # Run spec (test) file
 <cli> routine init              # Install built-in routines
+<cli> plugins list              # List registered apijack plugins
+<cli> plugins check             # Validate plugins (namespace, collisions, peer versions)
+```
+
+## Plugin System
+
+apijack supports pre-built plugins as standalone npm packages. Plugins register resolver functions under their own namespace (e.g., `@normalled/apijack-plugin-faker` exposes `$_faker(...)` for routines).
+
+### Installing a plugin
+
+```ts
+import { createCli } from "@apijack/core";
+import faker from "@normalled/apijack-plugin-faker";
+
+const cli = createCli({ name: "mycli", /* ... */ });
+cli.use(faker());              // zero-config
+cli.use(faker({ seed: 42 }));  // with default opts
+await cli.run();
+```
+
+### Per-routine plugin configuration
+
+```yaml
+name: seeded-user-gen
+plugins:
+  faker:
+    seed: 42
+steps:
+  - name: make-user
+    command: users create
+    args:
+      --name: "$_faker(person.fullName)"
+```
+
+Each routine invocation receives a fresh plugin state closure — routines are isolated from each other. Sub-routines without their own `plugins:` block inherit the parent's closures.
+
+### Plugin diagnostics
+
+```bash
+<cli> plugins list     # show installed plugins with version and status
+<cli> plugins check    # validate namespace, collision, and peer-version rules (exits non-zero on issue)
 ```
 
 ## Command Discovery with `-o routine-step`
--- a/README.md
+++ b/README.md
@@ -73,6 +73,10 @@ See [Building a CLI](https://github.com/normalled/apijack/wiki/Building-a-CLI) f
 - **Project mode** -- project-local config, routines, and extensions ([details](https://github.com/normalled/apijack/wiki/Project-Mode))
 - **OpenAPI 3.0 + 3.1** -- comprehensive spec support ([compatibility matrix](https://github.com/normalled/apijack/wiki/OpenAPI-Compatibility))
 
+## Plugins
+
+apijack supports pre-built plugins as standalone npm packages for common utilities. Install via `cli.use(plugin())` and configure per-routine via a top-level `plugins:` block in routine YAML. See `CLAUDE.md` for the full contract and `<cli> plugins list` to inspect registered plugins.
+
 ## Documentation
 
 Full documentation is available on the [wiki](https://github.com/normalled/apijack/wiki):
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -17,6 +17,7 @@ import { buildDispatcher } from './routine/dispatcher';
 import { homedir } from 'os';
 import { resolve, join } from 'path';
 import { registerPluginCommand } from './plugin/register';
+import { registerPluginsCommand } from './commands/plugins/register';
 import { PluginRegistry } from './plugin/registry';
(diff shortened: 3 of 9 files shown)


garretpremo (commented):
The new `plugins list`/`plugins check` commands are wired up, tested, and CI is green, but the new command group isn't integrated into two sibling wirings in `cli-builder.ts`, which affects real user flows.

**Blocking:**
- `src/cli-builder.ts:231` — `skipAuthCommands` does not include `'plugins'`. A user running `<cli> plugins list` or `<cli> plugins check` in a TTY without configured credentials will hit the interactive setup prompt (`Environment name: ...`), because `process.argv[2] === 'plugins'` is not in the skip set. Diagnostic commands should not require API auth — in fact, CI and plugin-author debugging scenarios are the primary case where creds may not be configured yet. Add `'plugins'` alongside the existing `'plugin'` entry.
- `src/cli-builder.ts:56` — `CORE_COMMANDS` does not include `'plugins'`. `showCustomHelp` uses this set to separate core commands from API command groups, so `plugins` gets counted under "API Commands" (e.g., "17 API command groups available" when one of them is the built-in diagnostic). Add `'plugins'` next to `'plugin'`.

<details>

- Spec gap in `plugins list` (`docs/superpowers/specs/2026-04-23-prebuilt-plugins-design.md` §Diagnostics): the spec prescribes columns `NAME | VERSION | PEER | STATUS` with STATUS values `ok | peer-mismatch | namespace-violation | collision`. The PR implements `NAME | VERSION | STATUS` with STATUS hardcoded to `'ok'`. In practice, invalid plugins can't even reach `list` because `cli.run()` throws at startup — but the missing PEER column still has user value (lets consumers verify peer ranges at a glance). Worth either updating the implementation or explicitly amending the spec.
- `plugins check` only reports the first error. `PluginRegistry.validateAll()` throws on the first failure, so `check.ts` catches one error and exits. The spec describes "a report of all issues grouped by plugin"; the current shape forces users to fix and re-run repeatedly when they have multiple problems. Consider accumulating errors in `validateAll()` or iterating plugins in `check.ts` itself.
- `plugins check` is effectively redundant with startup. Because `cli.run()` runs the exact same `validateAll()` + peer-range check before Commander executes the action, if validation would fail, the startup throw fires before `check`'s action ever runs; if startup succeeds, `check` always prints `All plugins OK.`. The CI use case still works via exit codes, but the "clean diagnostic report" value in the spec isn't realized. One option: have `plugins check` do its own non-throwing walk, decoupled from startup, so it can produce a clean report even when startup would crash.
- `PluginPeerMismatchError` is caught in `check.ts` but `validateAll()` never throws it — peer validation lives in `cli-builder.ts`, outside the registry. Either move peer validation into the registry (so `check` actually covers peers) or drop the dead import.
- Spec §"MCP exposure" claims both commands appear as MCP tools automatically. The current MCP pipeline (`src/mcp/tools/**`) requires explicit tool definitions per command (see `src/mcp/tools/config/config-switch/`). This PR adds none. If MCP coverage was a v1.9.0 goal, it's deferred here — fine to punt but worth calling out so it isn't lost.

</details>

<details>
<summary>Nitpicks</summary>

- `_coreVersion` is unused in both `src/commands/plugins/list.ts` and `src/commands/plugins/check.ts`. Drop the parameter (and the plumbing through `registerPluginsCommand`) or start using it.
- `check.ts` writes error messages to `process.stdout`; stderr is more conventional for diagnostic failures.
- `tests/commands/plugins.test.ts` hardcodes `'1.9.0'` as the core version argument. Since the param is unused today it doesn't matter, but if you keep it, consider reading `package.json` or using a sentinel.

</details>

> Note: posted as a comment instead of --request-changes because GitHub doesn't allow requesting changes on your own PR. The `changes requested` label reflects the outcome.

Reviewed by claude 🤖

garretpremo (commented):
The blocking items from the previous review are addressed in 95a16ce — `'plugins'` is now in both `CORE_COMMANDS` and `skipAuthCommands`, errors accumulate via `validateAllCollected`, the PEER column is added to `list`, and `cli.run()` skips throwing validation when invoked as `plugins check` so the diagnostic actually gets to run. Tests pass locally (829/0), CI is green, and lint is clean on touched files.

<details>

- **Brittle argv sniffing** (`src/cli-builder.ts:158`) — `process.argv[2] === 'plugins' && process.argv[3] === 'check'` works today but breaks if a global flag is ever added before the subcommand (e.g. `<cli> --verbose plugins check`). A more robust pattern would attach the non-throwing guard to the `plugins check` action itself (preAction hook) rather than positional argv inspection.
- **Peer-version validation duplicated** (`src/cli-builder.ts:170-184` vs `src/commands/plugins/check.ts:22-36`) — the same loop over `plugin.__package` + `loadPluginPeerInfo` + `checkPeerRange` lives in two places. Consider hoisting it onto `PluginRegistry` (alongside `validateAllCollected`) so `check` truly covers everything `cli.run()` does.
- **Silent skip of plugins missing `__package`** in `check.ts` — `cli.run()` writes a stderr warning when a plugin doesn't self-report its package; `plugins check` silently skips. For a CI-oriented diagnostic that's the more important place to surface it. Consider emitting the same warning (and possibly making it an error under a `--strict` flag).
- **MCP exposure still deferred** — the spec (`docs/superpowers/specs/2026-04-23-prebuilt-plugins-design.md` §MCP exposure) calls for `plugins list` / `plugins check` to appear as MCP tools. This PR doesn't add them under `src/mcp/tools/`. Fine to punt to a follow-up but worth tracking so it doesn't get lost when v1.9.0 ships.

</details>

<details>
<summary>Nitpicks</summary>

- `tests/commands/plugins.test.ts` hardcodes `'1.9.0'` as the core-version arg in every `registerPluginsCommand` call. If you ever bump the version, these strings drift; reading from `package.json` (or a constant) would keep them tied to reality.
- `check.ts` reports `Found N issue(s).` to stderr on failure — the parenthesized plural is fine but `Found N issues.` reads slightly cleaner (and `N === 1` vs `N > 1` is rare enough not to matter).

</details>

Reviewed by claude 🤖

Issue #48, opened 2026-04-25
Integration tests for boot-time cli.use / cli.command / cli.dispatcher loops in bin/apijack.ts


garretpremo:
Spawned from #46.

## Context

In the review of #46 (`feat/plugins-loader`), the reviewer noted that loader-level tests are thorough, but the actual `bin/apijack.ts` registration loops — the part that calls `cli.use(plugin)`, `cli.command(...)`, `cli.dispatcher(...)`, `cli.resolver(...)` — have no automated coverage. That code path is only exercised by a manual smoke test described in the PR body.

This is **consistent across the whole `.apijack/*` extension family** (auth, commands, dispatchers, resolvers, plugins) — none of them have integration coverage of the boot-time wire-up. So this is a family-level gap, not a plugins-specific one.

## Proposed work

Add a small integration test (or test suite) that:

1. Lays out a temp project with `.apijack.json` + `.apijack/{auth,commands/foo,dispatchers/bar,resolvers/_baz,plugins}.ts`
2. Boots the binary in that project (via `spawn` or by importing/calling the bootstrap directly)
3. Asserts each registered extension is reachable through its expected runtime surface — e.g., `<cli> plugins list` shows the plugin, `<cli> foo` runs, the dispatcher is invocable from a routine, `$_baz()` resolves in a routine

This closes the gap for all five extension types in one pass and gives any future loader (e.g., a hypothetical `generators/`) a template to follow.

## Out of scope

- Refactoring the loaders themselves (their unit tests already cover their contracts)
- Cross-extension dependency tests (e.g., a project resolver that wraps a plugin-provided function — that's a separate test concern)

## References

- Reviewer comment: https://github.com/normalled/apijack/pull/46#pullrequestreview-...  (under "Non-blocking observations" → bullet 2)
- Sibling loaders with the same gap: `loadProjectAuth`, `loadProjectCommands`, `loadProjectDispatchers`, `loadProjectResolvers`, `loadProjectPlugins` in `src/project-loader.ts`

PR #49, opened 2026-04-25
chore: triage cron jobs, review-body helper, dev-workflow consolidation

garretpremo:
## Summary

Adds the cron-job halves of the triage workflow (internal allowlisted + public sanitized paths), tightens up review-posting to avoid backtick-escape regressions, fixes a release-notes regression in `publish.yml`, and consolidates the redundant `dev-workflow` skill into the `triage-issue` → `implement-issue` chain.

The triage cron jobs land disabled — internal will be flipped on after a manual dry-run, public stays off until a stage-2 validator is in place. Both paths defense-in-depth lock + snapshot the issue at job start and verify before any write so post-hoc author edits are caught (GitHub does not prevent issue authors from editing their own bodies).

### Triage cron jobs (`111d906`)

- `.claude-jobs/triage-internal.yaml` (disabled) — gates by author allowlist, locks + snapshots, verifies before handoff. Handoff is label-only via `set-triage-label.sh ready-for-implement`; never invokes `Skill(implement-issue)` from cron.
- `.claude-jobs/triage-public.yaml` (disabled) — gates to non-allowlisted authors, runs `sanitize-issue.sh` (strips zero-width / unicode-tag / bidi chars, HTML comments, markdown image refs). Strictly read-shaped tool surface; only writable target is `.claude-jobs/triage-reports/<n>.md` (gitignored). System prompt teaches posture (treat content as data, refuse encoded payloads, refuse meta-instructions) without enumerating attack patterns.
- Helper scripts under `.claude/skills/triage-issue/scripts/`: `set-triage-label.sh`, `add-triage-flag.sh`, `ensure-triage-labels.sh`, `lock-issue.sh`, `snapshot-issue.sh`, `verify-snapshot.sh`, `sanitize-issue.sh`.
- `triage-issue` SKILL.md updated to call lock + snapshot at step 1 and verify before handoff; documents the cron-mode override (label-only handoff).
- `.gitignore` extended to keep snapshots / triage reports / the operator-private threat-model doc local-only.

### Review-posting helper (`17f3e10`, `f6747a0`)

Inline `gh pr review --body "$(cat <<'EOF'...EOF)"` repeatedly produced literally-escaped backticks in posted reviews (model habit from shells where backticks need escaping). Symptom: every code-fence inline like `` `.apijack.json` `` showed up as `` \`.apijack.json\` `` in #48.

- New `scripts/post-review.sh <pr> <decision> <body-file>` calls `gh pr review --body-file`. Path check: rejects any body path containing `..` (prevents traversal through the prefix glob), then enforces the `.claude-jobs/review-bodies/` prefix.
- `review-issue` SKILL.md: write the body to `.claude-jobs/review-bodies/<pr>.md` with `Write` (literal storage, no shell context, no escaping), then call the helper.
- `review-issue.yaml` allows `Write` (the path-check confines the blast radius).
- One-line note in `CLAUDE.md` so ad-hoc `gh issue/pr` calls also use `--body-file`.
- `.claude-jobs/review-bodies/` is gitignored.

### Release-notes regression fix (`62f9383`)

`publish.yml` was regenerating release notes from `git log` after the merge. That clobbered the curated PR body produced by `ship-release`. Switched to using the merged PR body as-is.

### Skill consolidation (`64ca121`, `f6747a0`)

The standalone `dev-workflow` skill duplicated what `triage-issue` + `implement-issue` already cover for issue-driven work. Two near-parallel docs were drifting.

- Deleted `.claude/skills/dev-workflow/`.
- `implement-issue` updated to absorb dev-workflow's recent additions:
  - Worktree as preferred branching method, with the "if `dev` is behind `main`, sync first" sequence (and a note that the resulting `git push origin dev` is the only push to `dev` this skill performs).
  - PR body format (Summary / What changes / Acceptance criteria / Test plan / Spec reference) — written for the automated reviewer.
  - `./scripts/wait-for-review.sh <pr>` invoked in background, replacing `gh pr checks --watch`.
  - New step 9: address-feedback loop — categorize blocking/non-blocking, fix-or-file-followup, push, re-invoke wait, loop.
  - STOP fires on `first pass reviewed` (was: PR-open + CI green). Merging is still `review-issue`'s job.
- `ship-release` cross-reference: `dev-workflow` → `triage-issue` / `implement-issue`.

## Acceptance criteria

- [x] Both triage cron job YAMLs ship `enabled: false`; flipping requires a follow-up
- [x] `verify-snapshot.sh` exits non-zero when the issue body has been edited since the snapshot, and the cron's `add-triage-flag.sh` path applies `edited-during-triage` and stops
- [x] `post-review.sh` rejects any `<body-file>` path outside `.claude-jobs/review-bodies/`, **including paths that traverse through `..`** (verified locally)
- [x] `dev-workflow` skill no longer exists and no remaining file references it
- [x] `implement-issue` STOP requires `first pass reviewed` verdict, not "PR open + CI green"
- [x] `ship-release` cross-reference points at `triage-issue` / `implement-issue`

## Test plan

- [x] `grep -rn "dev-workflow"` across the repo — 0 hits after the consolidation commit
- [x] Triage helper scripts are executable
- [x] `post-review.sh` `..` rejection: `bash post-review.sh 49 comment '.claude-jobs/review-bodies/../README.md'` → exits 2 with "body file path must not contain '..'"
- [x] `post-review.sh` prefix rejection: same script with `/etc/passwd` → exits 2 with prefix error
- [ ] Dry-run of `triage-internal.yaml` against a test issue (manual; happens after this PR lands)

## Follow-ups

Filed from the first review pass (out-of-scope refactors / nits):

- #50 — `sanitize-issue.sh`: replace bracket-pair regex with `jq -s 'add'` for pagination
- #51 — triage cron yamls: single source for `INTERNAL_AUTHORS` allowlist
- #52 — triage label helpers: don't print "added" when label was already present
- #53 — extract shared canonical-fields template for `snapshot-issue.sh` / `verify-snapshot.sh`


garretpremo (commented):

The triage cron jobs, review-body helper, and `dev-workflow` consolidation are well-scoped and the threat-modeling around the public/internal split is thoughtful. One acceptance criterion is **not strictly met**: the path check in `post-review.sh` accepts traversal through `..`, so paths that resolve outside `.claude-jobs/review-bodies/` are not actually rejected. Trivial fix; once that's tightened the PR looks good to merge.

**Blocking:**

- **AC #3 not strictly met — path traversal in `.claude/skills/review-issue/scripts/post-review.sh`.** The check is `case "$body" in .claude-jobs/review-bodies/*) ;; ...` — bash case-pattern `*` matches any string including `/` and `..`, so a body argument like `.claude-jobs/review-bodies/../README.md` (or any `../...`) passes the prefix check and is then read by `gh pr review --body-file`. The AC explicitly says "rejects any `<body-file>` path outside `.claude-jobs/review-bodies/`". Practical risk is bounded (the cron yaml grants a broader `Write` capability anyway, as the comment acknowledges), but the defense-in-depth check should at least do what the AC says. Suggested fix — one line, e.g.:

  ```bash
  case "$body" in
      *..*) echo "body file path must not contain '..'" >&2; exit 2 ;;
      .claude-jobs/review-bodies/*) ;;
      *) echo "body file must live under .claude-jobs/review-bodies/ (got: $body)" >&2; exit 2 ;;
  esac
  ```

  Or use `realpath` to canonicalize and re-check the prefix against the absolute `.claude-jobs/review-bodies/` directory.

<details>

- **`sanitize-issue.sh` paginated-comments parsing is fragile.** The regex `re.findall(r'\[.*?\](?=\s*\[|\s*$)', comments_raw, re.DOTALL)` is intended to split `gh api --paginate`'s concatenated JSON arrays, but `\[.*?\]` will also match any nested bracket pair inside comment payloads (e.g. a comment body containing `[link](url)` text serialized into the response). The single-page fallback rescues most real cases, but for issues with many comments containing brackets the extracted "chunks" can be malformed and silently dropped by the `try: json.loads(chunk) except: pass`. Cleaner: pipe through `jq -s 'add'` (as `snapshot-issue.sh` already does) or use `gh api --paginate ... --jq '.'` to flatten.

- **Scope: commit `62f9383` (publish.yml release-notes fix) isn't called out in the PR body's "What changes" section.** It's a small, well-targeted fix and tangentially related (release-notes generation), but it's a fourth bucket on top of the three the PR title and "What changes" list. Worth a one-line mention in the description so the diff matches the narrative, or split into its own PR for cleaner history.

- **`triage-public.yaml` lock + snapshot run inside `prompt_cmd`, which is harness-shell, not agent-tool.** That's intentional and correct (so the snapshot is taken before the prompt is even constructed), but the `allowed_tools` list won't gate those calls — worth a one-line comment in the yaml so a future reader doesn't try to remove `gh api*` from `allowed_tools` thinking it's used by `prompt_cmd`.

</details>

<details>
<summary>Nitpicks</summary>

- `triage-internal.yaml` defines `INTERNAL_AUTHORS=("garretpremo")` in both `preflight` and `prompt_cmd`. As the allowlist grows this is a synchronization hazard. A single source (env var on the job, or a shared shell snippet) would avoid the drift.

- `add-triage-flag.sh` `flags=()` array and `set-triage-label.sh` `state_labels=()` array each silently allow re-application of an already-present label; that's fine, but the printed "added" message is misleading when the label was already there. (Compare `lock-issue.sh` which explicitly handles "already locked".)

- `verify-snapshot.sh` recomputes the snapshot canonical hash inline rather than calling a shared function with `snapshot-issue.sh`. If the canonical-fields list ever changes in one place but not the other, the verifier silently starts failing for unrelated reasons. Extracting the `jq -n '{...}'` template to a shared file would prevent the drift.

- `implement-issue` SKILL.md step 1 instructs `git merge origin/main` into `dev` and `git push origin dev` from a feature-branch flow — this writes to `dev` from the implementation skill, which is otherwise scoped to feature branches. Worth either gating that on a separate manual step or noting that this `dev`-sync push is the one explicit exception.

</details>

garretpremo (commented):

Re-review after the prior `changes requested` verdict. The blocking AC #3 — `..` traversal in `.claude/skills/review-issue/scripts/post-review.sh` — is now strictly enforced (the `*..*) ... exit 2` arm is the first match in the `case`, so a path like `.claude-jobs/review-bodies/../README.md` is rejected before the prefix check runs). All six acceptance criteria are met, CI is green across `lint`, `unit-tests`, and the e2e matrix, and the prior non-blocking observations and nits have either been folded in inline (publish.yml call-out in the PR body, the `prompt_cmd`-vs-`allowed_tools` comment in `triage-public.yaml`, the `dev`-sync push exception note in `implement-issue` SKILL.md) or filed as follow-ups (#50–#53). Looks good to merge.

<details>

- The `*..*` guard is intentionally conservative — it would also reject a legitimate filename like `foo..bar.md`. Harmless in practice (the only producer is the skill itself, which writes `.claude-jobs/review-bodies/<pr>.md`), but worth knowing if anyone later wants to broaden the convention.

- `snapshot-issue.sh` and `verify-snapshot.sh` still each inline the canonical-fields jq object literal. By inspection the two are byte-identical and produce the same `jq -Sc | sha256sum` output, so the verifier won't drift today; #53 is the right place to defuse the future drift risk.

</details>

PR #62, opened 2026-04-26
fix: generate command honors project AuthStrategy (#61)

### LINKED ISSUE #61: generate command should honor project AuthStrategy, not just Basic auth
## Problem

`<cli> generate` is hardcoded to HTTP Basic auth and ignores the project's `AuthStrategy`. For any API behind a session/cookie/2FA flow, this means the spec fetch returns 401 even though the rest of the CLI authenticates fine — so the project must shadow `generate` with its own implementation just to drive the existing `AuthStrategy` against `/v3/api-docs`.

Concrete example: rrc-cli (`@apijack/core` consumer, RRCloud has Spring Security + 2FA) ships a hand-written [`commands/generate.ts`](https://github.com/Rational-Retention/rrc-cli/blob/development/.apijack/commands/generate.ts) (~75 lines) that loads `loadProjectAuth`, calls `SessionManager.resolve(strategy, auth)`, then re-implements the spec fetch with `{ Accept: 'application/json', ...session.headers }`. None of those 75 lines are project-specific — they're the work core's `generate` would do if it knew about the strategy.

## Where it breaks

`src/codegen/index.ts` `fetchAndGenerate` only knows Basic:

```ts
export async function fetchAndGenerate(opts: FetchAndGenerateOptions): Promise<void> {
    const headers: Record<string, string> = { Accept: 'application/json' };
    if (opts.auth) {
        headers.Authorization = 'Basic ' + btoa(`${opts.auth.username}:${opts.auth.password}`);
    }
    const url = `${opts.baseUrl}${opts.specPath}`;
    // ...
}
```

`src/commands/generate/generate.ts` calls it with bare username/password:

```ts
await input.fetchAndGenerate({
    baseUrl: input.env.url,
    specPath: input.specPath,
    outDir: input.outDir,
    auth: { username: input.env.user, password: input.env.password },
});
```

`src/cli-builder.ts:223` registers the `generate` command **before** `strategy` and `sessionMgr` are computed (line 291+) and never threads them in:

```ts
registerGenerateCommand(program, cliName, options.specPath, generatedDir, configOpts);
// ...later, line 291+
let strategy = options.auth;       // ← project AuthStrategy (already resolved by bin/apijack.ts)
let sessionMgr: SessionManager | null = null;
```

`options.auth` already carries the project's strategy (see `bin/apijack.ts:86-102` — `loadProjectAuth` runs first and feeds it to `createCli({ auth })`), so the data is right there. It just doesn't reach `generate`.

## Approaches considered

Brainstormed three directions; selected the third. Tradeoffs documented for posterity / so reviewers can push back.

### Approach 1 — Additive params on `fetchAndGenerate`

Add optional `strategy?: AuthStrategy` + `sessionManager?: SessionManager` to `fetchAndGenerate`'s opts; branch internally on whether they're present.

- **Pro:** smallest possible diff; one function still does the whole job; one call site to update.
- **Con:** piles auth-mode-selection branching into a low-level utility; soft-deprecate signal is fuzzy because the function gets *more* capable, not less.

### Approach 2 — CLI command owns the fetch; codegen stays pure

Move spec-fetch (and 401-retry) into `generateAction` in `src/commands/generate/generate.ts`. CLI passes parsed spec to the already-exported pure `generate({ spec, outDir })`. Leave `fetchAndGenerate` untouched, soft-deprecate via JSDoc.

- **Pro:** clean separation (CLI = I/O + auth, codegen = transform).
- **Con:** `fetchAndGenerate` and the new CLI path end up with two implementations of "fetch a spec"; the auth + 401-retry logic isn't independently testable — only reachable through `generateAction`.

### Approach 3 — Extract `fetchSpec` helper, share between both ✅ selected

New exported `fetchSpec({ baseUrl, specPath, strategy?, sessionManager?, auth? })` returning parsed spec JSON, with the once-on-401 invalidate-and-retry baked in. `fetchAndGenerate` becomes `fetchSpec → generate`. CLI calls `fetchSpec` directly with strategy, then `generate({ spec, outDir })`.

- **Pro:**
  - Testable in isolation (mock `fetch`, assert retry-and-invalidate sequence).
  - Single source of truth for spec-fetch — `fetchAndGenerate` and the CLI both go through it.
  - Self-documenting: "spec fetching happens in one named place."
- **Con:** net +1 exported function on the public API surface; ~20 extra lines for the helper + tests vs. inlining.

**Why 3 over 2:** the `fetchSpec` extraction isn't speculative — the CLI has to fetch the spec somehow, and the only question is whether that logic is inline or named/exported. A named helper for auth + retry is exactly the kind of code that benefits from isolated tests, and the cost (one extra exported symbol) is small.

## Proposed fix (Approach 3)

```ts
// src/codegen/index.ts — new export
export interface FetchSpecOptions {
    baseUrl: string;
    specPath: string;
    auth?: { username: string; password: string };
    strategy?: AuthStrategy;
    sessionManager?: SessionManager;
}

export async function fetchSpec(opts: FetchSpecOptions): Promise<unknown> {
    const buildHeaders = async (): Promise<Record<string, string>> => {
        const headers: Record<string, string> = { Accept: 'application/json' };
        if (opts.strategy && opts.sessionManager && opts.auth) {
            const session = await opts.sessionManager.resolve(opts.strategy, {
                baseUrl: opts.baseUrl,
                username: opts.auth.username,
                password: opts.auth.password,
            });
            Object.assign(headers, session.headers);
        } else if (opts.auth) {
            headers.Authorization = 'Basic ' + btoa(`${opts.auth.username}:${opts.auth.password}`);
        }
        return headers;
    };

    const url = `${opts.baseUrl}${opts.specPath}`;
    let res = await fetch(url, { headers: await buildHeaders() });

    // Cached session may be stale — invalidate + retry once on 401
    if (res.status === 401 && opts.strategy && opts.sessionManager) {
        opts.sessionManager.invalidate();
        res = await fetch(url, { headers: await buildHeaders() });
    }

    if (!res.ok) throw new Error(`Failed to fetch spec from ${url}: ${res.status} ${res.statusText}`);
    return res.json();
}

// fetchAndGenerate becomes a thin composition (signature unchanged)
export async function fetchAndGenerate(opts: FetchAndGenerateOptions): Promise<void> {
    const spec = await fetchSpec(opts);
    await generate({ spec, outDir: opts.outDir });
}
```

CLI wiring (`src/commands/generate/generate.ts` + `src/cli-builder.ts`): extend `GenerateInput` with optional `strategy` + `sessionManager`; `generateAction` calls `fetchSpec` then `generate`. Reorder cli-builder so auth resolves before `registerGenerateCommand` runs (or thread the values via a thunk to keep current order).

## Out of scope (worth flagging separately)

The once-on-401 invalidate-and-retry pattern is general — any cached-session consumer wants it. Belongs in `SessionManager.resolve` eventually, but not this ticket. For now, `fetchSpec` owns it.

## Files to touch

- `src/codegen/index.ts` — add `fetchSpec` export; refactor `fetchAndGenerate` to compose it
- `src/codegen/index.test.ts` (new or existing) — unit tests on `fetchSpec`: Basic-only path, strategy path, 401-invalidate-and-retry, retry-still-fails
- `src/commands/generate/generate.ts` — extend `GenerateInput`; `generateAction` calls `fetchSpec` + `generate`
- `src/commands/generate/generate.test.ts` — update mocks for new inputs
- `src/cli-builder.ts:223` — pass `strategy` + `sessionMgr` to `registerGenerateCommand` (reorder or use a thunk)

## Outcome

`rrc-cli` (and any other 2FA / session-gated consumer) deletes its `commands/generate.ts` shadow.

garretpremo:
Closes #61

## Summary

`generate` was hardcoded to HTTP Basic auth and ignored the project's `AuthStrategy`. For any API behind a session/cookie/2FA flow, this meant the spec fetch returned 401 even though the rest of the CLI authenticated fine — so consumers had to shadow `generate` with custom code (e.g., rrc-cli ships ~75 lines of boilerplate).

This PR follows Approach 3 from the issue: extract a new exported `fetchSpec` helper that handles both Basic and session-strategy paths, with once-on-401 invalidate-and-retry. `fetchAndGenerate` becomes a thin `fetchSpec → generate` composition (public signature unchanged). Strategy + sessionManager are threaded through `GenerateInput` and `registerGenerateCommand`. The cli-builder is reordered so `generate` registers after `strategy`/`sessionMgr` are computed.

### `src/codegen/index.ts`

- New exported `FetchSpecOptions` interface and `fetchSpec` function
- `fetchSpec` builds headers from either `strategy + sessionManager + auth` (resolves a session, copies its headers) or falls back to Basic from `auth`
- On a 401 with strategy + sessionManager, calls `sessionManager.invalidate()` and retries once
- `fetchAndGenerate` becomes `fetchSpec → generate`; signature unchanged

### `src/commands/generate/generate.ts`

- `GenerateInput` now exposes `strategy?`, `sessionManager?`, and separate `fetchSpec` + `generate` injection points (replacing the single `fetchAndGenerate` injection)
- `generateAction` calls `fetchSpec` then `generate` directly
- `registerGenerateCommand` accepts optional `strategy` and `sessionManager` and threads them through

### `src/cli-builder.ts`

- `registerGenerateCommand` moved from step 3 (before auth resolution) to a new step 5b (after `strategy` and `sessionMgr` are computed). Registration just defines the action callback — execution still happens at `parseAsync()` time, so the reorder is purely about thread-through, not behavior.

### Tests

- New `src/codegen/index.test.ts`: Basic-only path, strategy path, 401-invalidate-retry success, 401-retry-still-fails, no-retry-without-strategy
- Updated `src/commands/generate/generate.test.ts`: new injection shape, plus a test that strategy + sessionManager are threaded to fetchSpec

## Acceptance criteria

- [x] New exported `fetchSpec` accepts `auth`, `strategy`, `sessionManager`
- [x] 401 triggers a single `sessionManager.invalidate()` + retry
- [x] `fetchAndGenerate` keeps its public signature; becomes `fetchSpec → generate`
- [x] `src/codegen/index.test.ts` covers Basic, strategy, 401-retry-success, 401-retry-fail
- [x] `generateAction` calls `fetchSpec` then `generate`, threading optional `strategy` + `sessionManager`
- [x] `cli-builder.ts` passes `strategy` + `sessionMgr` to `registerGenerateCommand`
- [x] Existing `generate.test.ts` updated for new inputs
- [x] A 2FA/session-gated consumer (e.g., rrc-cli) can drop its `commands/generate.ts` shadow

## Test plan

- [x] `bun test` — 841 pass / 0 fail
- [x] `bun run lint` — 0 errors on changed files
- [x] Independent code review (subagent) — no Critical/Important issues; verdict ready to commit

## Backwards compatibility

- `fetchAndGenerate` keeps its exact public signature
- `GenerateInput` is internal (not part of the public API surface)
- Generated code paths unchanged

### DIFF
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -220,7 +220,6 @@ export function createCli(options: CliOptions): Cli {
                 knownSites: options.knownSites,
                 allowedCidrs: options.allowedCidrs,
             });
-            registerGenerateCommand(program, cliName, options.specPath, generatedDir, configOpts);
             registerUpgradeCommand(program, options.version);
             registerMcpCommand(
                 program,
@@ -303,6 +302,17 @@ export function createCli(options: CliOptions): Cli {
                     : options.auth;
                 sessionMgr = new SessionManager(cliName, join(configDir, 'session.json'));
             }
+
+            // 5b. Register generate (depends on strategy + sessionMgr from step 5)
+            registerGenerateCommand(
+                program,
+                cliName,
+                options.specPath,
+                generatedDir,
+                configOpts,
+                strategy,
+                sessionMgr,
+            );
 
             // 6. Import generated commands (ALWAYS — no auth needed)
             let commandsModule: Record<string, unknown> | null = null;
--- a/src/codegen/index.test.ts
+++ b/src/codegen/index.test.ts
@@ -1,0 +1,137 @@
+import { describe, test, expect, mock, beforeEach, afterEach } from 'bun:test';
+import { fetchSpec } from './index';
+import type { AuthStrategy, AuthSession } from '../auth/types';
+import type { SessionManager } from '../session';
+
+const stubResponse = (status: number, body: unknown = {}) =>
+    new Response(JSON.stringify(body), { status });
+
+describe('fetchSpec', () => {
+    const originalFetch = globalThis.fetch;
+    let fetchMock: ReturnType<typeof mock>;
+
+    beforeEach(() => {
+        fetchMock = mock(() => Promise.resolve(stubResponse(200, { paths: {} })));
+        globalThis.fetch = fetchMock as unknown as typeof fetch;
+    });
+
+    afterEach(() => {
+        globalThis.fetch = originalFetch;
+    });
+
+    test('uses Basic auth when no strategy is provided', async () => {
+        await fetchSpec({
+            baseUrl: 'http://localhost:8080',
+            specPath: '/v3/api-docs',
+            auth: { username: 'admin', password: 'secret' },
+        });
+
+        expect(fetchMock).toHaveBeenCalledTimes(1);
+        const [url, init] = fetchMock.mock.calls[0] as [string, RequestInit];
+        expect(url).toBe('http://localhost:8080/v3/api-docs');
+        const headers = init.headers as Record<string, string>;
+        expect(headers.Authorization).toBe(
+            'Basic ' + btoa('admin:secret'),
+        );
+        expect(headers.Accept).toBe('application/json');
+    });
+
+    test('uses session headers from strategy when provided', async () => {
+        const session: AuthSession = {
+            headers: { Cookie: 'JSESSIONID=abc; X-XSRF=xyz' },
+        };
+        const sessionManager = {
+            resolve: mock(() => Promise.resolve(session)),
+            invalidate: mock(() => {}),
+        } as unknown as SessionManager;
+        const strategy = {} as AuthStrategy;
+
+        await fetchSpec({
+            baseUrl: 'http://localhost:8080',
+            specPath: '/v3/api-docs',
+            auth: { username: 'admin', password: 'secret' },
+            strategy,
+            sessionManager,
+        });
+
+        expect(sessionManager.resolve).toHaveBeenCalledWith(strategy, {
+            baseUrl: 'http://localhost:8080',
+            username: 'admin',
+            password: 'secret',
+        });
+        const [, init] = fetchMock.mock.calls[0] as [string, RequestInit];
+        const headers = init.headers as Record<string, string>;
+        expect(headers.Cookie).toBe('JSESSIONID=abc; X-XSRF=xyz');
+        expect(headers.Authorization).toBeUndefined();
+    });
+
+    test('on 401, invalidates session and retries once', async () => {
+        const session: AuthSession = { headers: { Cookie: 'fresh' } };
+        const resolve = mock(() => Promise.resolve(session));
+        const invalidate = mock(() => {});
+        const sessionManager = { resolve, invalidate } as unknown as SessionManager;
+        const strategy = {} as AuthStrategy;
+
+        let callCount = 0;
+        fetchMock = mock(() => {
+            callCount++;
+
+            return Promise.resolve(
+                callCount === 1
+                    ? stubResponse(401, {})
+                    : stubResponse(200, { paths: {} }),
+            );
+        });
+        globalThis.fetch = fetchMock as unknown as typeof fetch;
+
+        const spec = await fetchSpec({
+            baseUrl: 'http://localhost:8080',
+            specPath: '/v3/api-docs',
+            auth: { username: 'admin', password: 'secret' },
+            strategy,
+            sessionManager,
+        });
+
+        expect(invalidate).toHaveBeenCalledTimes(1);
+        expect(resolve).toHaveBeenCalledTimes(2);
+        expect(fetchMock).toHaveBeenCalledTimes(2);
+        expect(spec).toEqual({ paths: {} });
+    });
+
+    test('throws when retry after 401 still fails', async () => {
+        const sessionManager = {
+            resolve: mock(() => Promise.resolve({ headers: {} } as AuthSession)),
+            invalidate: mock(() => {}),
+        } as unknown as SessionManager;
+        const strategy = {} as AuthStrategy;
+
+        fetchMock = mock(() => Promise.resolve(stubResponse(401, {})));
+        globalThis.fetch = fetchMock as unknown as typeof fetch;
+
+        await expect(
+            fetchSpec({
+                baseUrl: 'http://localhost:8080',
+                specPath: '/v3/api-docs',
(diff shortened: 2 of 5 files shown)


garretpremo (commented):

Approach 3 from the issue is implemented faithfully. `fetchSpec` is exported, threads `strategy + sessionManager + auth`, and the once-on-401 invalidate-and-retry is in place. `fetchAndGenerate` keeps its public signature and is now `fetchSpec → generate`. `cli-builder` is reordered so `registerGenerateCommand` runs after step 5 has computed `strategy` / `sessionMgr` — registration still completes before `parseAsync()`, so the reorder is purely a thread-through. Tests cover all four paths called out in the AC plus a fifth (no-retry without strategy). Lint clean, linux CI green; windows jobs were still pending at review time but match the repo's normal stress-matrix lag.

<details>

- Subtle behavior change: when an env is resolved, `sessionMgr` is non-null even for plain Basic CLIs, so `generate` now goes through `SessionManager.resolve` and writes a `session.json` (just `{ headers: { Authorization: 'Basic …' } }`). Previously `generate` did inline Basic and never touched the session file. Benign — `invalidate-on-401` covers password rotation — and it makes `generate` consistent with the rest of the CLI, but worth flagging since it's a surface-level behavior diff that won't show up in a unit test.
- `buildHeaders` re-runs `sessionManager.resolve` on the retry path. That's intentional (after `invalidate()` the cache is gone, so `resolve` re-authenticates), but a one-liner comment on the retry block — e.g. `// resolve() will re-authenticate after invalidate()` — would save a future reader the trace.
- `sessionManager?: SessionManager | null` in `registerGenerateCommand` accepts both `null` and `undefined`, then the call site does `sessionManager ?? undefined`. Fine, but slightly redundant — picking one shape (`SessionManager | null` with `?? undefined` at the boundary, or `SessionManager | undefined` end-to-end) would tidy it up.

</details>

<details>
<summary>Nitpicks</summary>

- `generate.test.ts` casts `fetchSpec.mock.calls[0]` via `as unknown as [Record<string, unknown>]`. The other tests in the file use the cleaner `[opts]` destructure with a typed mock — minor consistency thing.

</details>

PR #94, opened 2026-05-03
test: integration coverage for SessionAuthStrategy stale-session refresh + retry (#77)

### LINKED ISSUE #77: SessionAuthStrategy: auto-refresh and retry once on stale-session 401/403
## Summary

When `SessionAuthStrategy`'s cached session expires (server-side idle timeout), the next API call ships stale cookies and the server returns 401 (or 403, in some Spring configs). Apijack surfaces the error to the caller instead of detecting "stale session" → re-bootstrapping `/session` → retrying the original request once.

## Repro (via `@apijack/core` against any session-cookie-based API)

1. Authenticate so the session is cached (`.apijack/session.json`).
2. Idle past the server-side session timeout.
3. Run any API command, e.g. `<cli> <some resource> delete <id>`.
4. Server returns 403:
   ```
   {"timestamp":"...","status":403,"error":"Forbidden","path":"/some/path/5"}
   ```

Workaround: run anything that triggers a fresh hit to `/session` (e.g. a custom command that re-resolves the session, or `clear-session` + retry). After that re-bootstraps the cookies, subsequent calls succeed for the rest of the idle window.

## Expected

`SessionAuthStrategy` should recover transparently:

- On 401 (and optionally 403, configurable) returned to a request that was authed via the cached session, re-call `/session` once to refresh cookies, then retry the original request.
- Cap at one retry to avoid loops.
- If the refresh itself fails, propagate the original error untouched.

## Notes

- Hit while bulk-deleting test resources via the MCP server's `run_commands` after a long pause; first call returned 403, subsequent calls in the same MCP session worked once a session-refreshing command re-bootstrapped cookies.
- The cached session `extends` implicitly during active use (server returns `Set-Cookie SESSION` on each response), so this only manifests after sustained inactivity — easy to miss in dev, painful when scripts pause between steps.
- This is distinct from the `watch` known-issue around STOMP session expiry; the gap exists for ordinary HTTP commands too.

## Acceptance criteria

- An integration test that simulates a stale-session 401, asserts `/session` is called exactly once, and asserts the original request is retried successfully.
- Behavior is opt-in or default-on but documented. Strategies that don't use `/session` are unaffected.
- Brief note in the auth strategy docs.

garretpremo:
Closes #77

## Summary

Closes a test + docs gap, not a code gap. The stale-session refresh + retry mechanism described by #77 already exists end-to-end: `SessionAuthConfig.refreshOn` → wired into the generated `ApiClient` via `cli-builder.ts:248-253` → on a status in `refreshOn`, the client invokes the refresh callback (which calls `sessionMgr.invalidate()` then `sessionMgr.resolve()`, ending in `SessionAuthStrategy.authenticate()` re-bootstrapping `/session`) and retries the original request once. The mechanism was undocumented and untested, so consumers (notably the RRC CLI that triggered the report) didn't know to opt in.

This PR adds the missing integration coverage and the missing docs. Per triage, behavior stays opt-in — no default-on change. Refresh-callback-throws semantics (the issue's "propagate the original error untouched") is intentionally deferred; the breadcrumb is in the test.

### New integration test — `tests/auth/session-auth-stale-retry.integration.test.ts`

Two tests that wire the real components (`SessionManager`, `SessionAuthStrategy`, `BasicAuthStrategy`, `resolveRequestHeaders`) and a runtime-generated `ApiClient` produced by `generateClient()` and dynamic-imported from a tmp `.ts` file. The only seam mocked is `globalThis.fetch`.

1. **Happy path**: pre-populates a stale `session.json` on disk, stages fetch to return 401 on the first `DELETE /admin/matters/5`, 200 on `/session`, then 200 on the retried `DELETE`. Asserts:
   - `/session` was called exactly once (`expect(sessionCount).toBe(1)`)
   - The original `DELETE` was made twice
   - The retry shipped the **refreshed** cookie, not the stale one (proves `getHeaders()` is re-resolved post-refresh)
2. **Opt-in proof**: same setup, `refreshOn` omitted from the config, single 401 propagates with no refresh and no retry. Pins the contract that strategies without `refreshOn` are unaffected.

### Docs

- New "Stale-session refresh and retry (opt-in)" subsection in `CLAUDE.md` under "Authentication Strategies", with a copy-pasteable `SessionAuthStrategy` config snippet.
- JSDoc on `SessionAuthConfig.refreshOn` (`src/auth/types.ts`) so it's discoverable in IDE intellisense.

### No production code changed

`src/auth/session-auth.ts`, `src/codegen/client.ts`, `src/cli-builder.ts`, `src/session.ts` are all untouched.

## Acceptance criteria

- [x] Integration test simulates a stale-session 401, asserts `/session` is called exactly once, and asserts the original request is retried successfully
- [x] Behavior is opt-in (and now documented). Strategies that don't use `/session` are unaffected
- [x] Brief note in the auth strategy docs

## Test plan

- [x] `bun test` — 868 pass / 0 fail
- [x] `bun run lint` — 0 errors (108 pre-existing warnings unchanged)
- [x] `bun test tests/auth/session-auth-stale-retry.integration.test.ts` — 2 pass / 0 fail
- [x] Self-review via `superpowers:code-reviewer` — verdict: ready to merge

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -202,6 +202,20 @@ apijack supports pluggable authentication:
 
 Generated OpenAPI commands resolve the session automatically. Consumer-registered custom commands and dispatchers opt in — see Project Extensions below.
 
+### Stale-session refresh and retry (opt-in)
+
+`SessionAuthStrategy`'s cached cookies can go stale on the server side (idle timeout) without the client knowing — the next call ships the dead cookie and the server returns 401 (or 403 in some Spring configs). To recover transparently, opt in by setting `refreshOn` on the session config:
+
+```ts
+new SessionAuthStrategy(new BasicAuthStrategy(), {
+  session: { endpoint: '/session' },
+  cookies: { extract: ['SESSION'], applyTo: ['POST', 'PUT', 'DELETE'] },
+  refreshOn: [401],   // or [401, 403]
+});
+```
+
+When the generated client receives a status in `refreshOn`, it calls the wired refresh callback — which invalidates the cached session and re-bootstraps `/session` — then retries the original request once. Capped at one retry. Strategies that don't use `/session` are unaffected (the field is ignored).
+
 ## Project Extensions
 
 The `.apijack/` directory at a project root is auto-loaded when the CLI runs inside that project:
--- a/src/auth/types.ts
+++ b/src/auth/types.ts
@@ -31,6 +31,12 @@ export interface SessionAuthConfig {
         toHeader: string;
         applyTo?: string[];
     }>;
+    /**
+     * HTTP statuses that trigger a one-shot session refresh + retry on the original request.
+     * Opt-in. Common value for stale-session recovery: `[401]` (or `[401, 403]` for servers
+     * that map an expired session to 403). The generated client invokes the refresh callback,
+     * which re-bootstraps `/session`, then retries the original request once.
+     */
     refreshOn?: number[];
     /** Called when the session endpoint returns a non-OK response. Return query params to retry, or null to give up. */
     onChallenge?: (status: number, body: string) => Promise<Record<string, string> | null>;
--- a/tests/auth/session-auth-stale-retry.integration.test.ts
+++ b/tests/auth/session-auth-stale-retry.integration.test.ts
@@ -1,0 +1,225 @@
+import { describe, test, expect, beforeEach, afterEach } from 'bun:test';
+import { mkdtempSync, writeFileSync, rmSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+import { generateClient } from '../../src/codegen/client';
+import { SessionManager } from '../../src/session';
+import { SessionAuthStrategy } from '../../src/auth/session-auth';
+import { BasicAuthStrategy } from '../../src/auth/basic';
+import { resolveRequestHeaders } from '../../src/auth/resolve-headers';
+import type { OpenApiOperation } from '../../src/codegen/openapi-types';
+import type { SessionAuthConfig, AuthSession, ResolvedAuth } from '../../src/auth/types';
+
+/**
+ * Integration test for stale-session refresh + retry (#77).
+ *
+ * Exercises the full pipeline that recovers from a server-side session timeout:
+ *   cached session loaded → API call → 401 → /session re-bootstrap → retry → 200
+ *
+ * This is the contract documented in CLAUDE.md under "Stale-session refresh and
+ * retry": a consumer opts in by setting `refreshOn` on `SessionAuthConfig`, and
+ * the generated client + cli-builder wiring transparently refresh and retry
+ * once on the configured statuses.
+ */
+describe('SessionAuthStrategy stale-session refresh + retry (integration)', () => {
+    let tmpDir: string;
+    let sessionPath: string;
+    let originalFetch: typeof globalThis.fetch;
+
+    const baseUrl = 'https://api.example.com';
+    const resolved: ResolvedAuth = { baseUrl, username: 'user', password: 'pass' };
+    const sessionConfig: SessionAuthConfig = {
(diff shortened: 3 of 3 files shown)


garretpremo (commented):

LGTM. Acceptance criteria from #77 are met and the "no production code changed" claim checks out — the refresh+retry mechanism already lives in `src/cli-builder.ts:248-253` (passing `mergedSessionAuth?.refreshOn` and a closure over `ctx.refreshSession` into the generated `ApiClient`) and the generated `client.ts` request loop at lines 229-247. The new integration test wires the real `SessionManager` / `SessionAuthStrategy` / `BasicAuthStrategy` / `resolveRequestHeaders` / runtime-generated `ApiClient` and only mocks `globalThis.fetch`, which is the right shape for an integration spec on this contract. CI is fully green (lint, unit-tests on both runners, OAS 3.0/3.1 + edge-case suites on both runners).

<details>

- **403 isn't exercised by either test.** Issue #77 explicitly calls out that some Spring configs map an expired session to 403, and the new docs in `CLAUDE.md` show `refreshOn: [401, 403]` as a recommended value. The mechanism in `client.ts` uses `refreshOn?.includes(res.status)` so 403 is structurally identical to 401, but a small additional case (or table-driven test) would lock the docs example to the test contract. Worth a quick follow-up — not a gap in this PR.
- **Refresh-callback-failure path uncovered.** Issue #77's "Expected" section bullet 3 ("If the refresh itself fails, propagate the original error untouched") is genuinely not covered here, and the PR body + the in-test breadcrumb (`tests/auth/session-auth-stale-retry.integration.test.ts:106-108`) explicitly defer it. That's a fine call given scope — but the deferral would be easier to track as its own follow-up issue rather than only a code comment, since the breadcrumb will rot the moment that block of test code is refactored.
- **Test duplication between the two cases.** The second test re-declares `paths`, re-imports `ApiClient`, and re-builds the fetch mock. Self-contained tests have real value here (each one reads top-to-bottom), so this is a judgment call — flagging it because a small `setupClient(refreshOn?)` helper would cut ~30 lines without hiding what each case proves.

</details>

<details>
<summary>Nitpicks</summary>

- The runtime `import('client.ts')` from a tmp dir relies on Bun running TS natively — worth a one-line comment for future readers, since under Node + ts-node this pattern would fail. (`tests/auth/session-auth-stale-retry.integration.test.ts:74-79` and `:175-180`.)
- The generated `ApiClient` constructor's third/fourth params (`onRefreshNeeded`, `refreshOn`) are typed inline at each `import` call site instead of importing the generated `ApiClient` type. Minor; type stays correct, just slightly verbose.

</details>

garretpremo (commented):

Approved.

- **403 isn't exercised by either test.** Issue #77 explicitly calls out that some Spring configs map an expired session to 403, and the new docs in `CLAUDE.md` show `refreshOn: [401, 403]` as a recommended value. The mechanism in `client.ts` uses `refreshOn?.includes(res.status)` so 403 is structurally identical to 401, but a small additional case (or table-driven test) would lock the docs example to the test contract. Worth a quick follow-up — not a gap in this PR.
- **Refresh-callback-failure path uncovered.** Issue #77's "Expected" section bullet 3 ("If the refresh itself fails, propagate the original error untouched") is genuinely not covered here, and the PR body + the in-test breadcrumb (`tests/auth/session-auth-stale-retry.integration.test.ts:106-108`) explicitly defer it. That's a fine call given scope — but the deferral would be easier to track as its own follow-up issue rather than only a code comment, since the breadcrumb will rot the moment that block of test code is refactored.
- **Test duplication between the two cases.** The second test re-declares `paths`, re-imports `ApiClient`, and re-builds the fetch mock. Self-contained tests have real value here (each one reads top-to-bottom), so this is a judgment call — flagging it because a small `setupClient(refreshOn?)` helper would cut ~30 lines without hiding what each case proves.

Issue #83, opened 2026-04-30
Honor projectConfig.name/description in bin/apijack.ts (display-only branding)


garretpremo:
## Problem

When a downstream CLI delegates to `@apijack/core/bin/apijack.ts` (the auto-discovery / `.apijack/`-extension-dir pattern), the resulting binary brands itself as `apijack` in all user-facing output — `--help` text, the setup banner, "Run '<cli> setup'" hints, and Commander's program name. There is no way for a consumer to override this short of writing a custom entry point that reproduces the bin's discovery logic (env loading, project config, plugins, commands, dispatchers, resolvers, settings — ~150 lines of glue).

Concretely, in `bin/apijack.ts`:

```ts
const cli = createCli({
    name: CLI_NAME,                                                      // hardcoded 'apijack'
    description: 'Jack into any OpenAPI spec and rip a full-featured CLI', // hardcoded
    version: VERSION,                                                    // core's package.json
    ...
});
```

`ProjectConfig` already declares `name?: string` (see `src/project.ts:6`) but the bin never reads it.

## Proposed scope (display only)

Wire two fields from `.apijack.json` through to `createCli` so the CLI presents itself with the consumer's branding. Keep storage paths, session keys, env-var prefixes, and the global routines dir untouched — those involve a separate refactor with real migration concerns and shouldn't ride along on this ticket.

### Changes

1. **`src/project.ts`** — extend `ProjectConfig`:
   ```ts
   export interface ProjectConfig {
       name?: string;
       description?: string;   // NEW
       version?: string;       // NEW (optional; useful when the consumer wants their own version string)
       specUrl?: string;
       generatedDir?: string;
       auth?: string;
       allowedCidrs?: string[];
   }
   ```

2. **`src/cli-builder.ts`** — add a separate `programName` to `CliOptions` and use it for *display only*:
   ```ts
   interface CliOptions {
       name: string;            // storage/session/env-prefix identity (unchanged semantics)
       programName?: string;    // NEW — defaults to `name`; used for help, banner, program.name()
       description: string;
       version: string;
       // ...
   }
   ```
   Resolve `const displayName = options.programName ?? options.name` once, then thread `displayName` into:
   - `program.name(displayName)` (currently L194)
   - help-output strings (currently L117, L121, "Run '<cli> --help'…")
   - setup banner / "Saved environment to ~/.{cliName}/config.json" log (currently L259, L280)
   - "Not authenticated. Run '<cli> setup'…" error (currently L347)

[...]

PR #96, opened 2026-05-03
fix: thread displayName through subcommand registrar hints (#91)

### LINKED ISSUE #91: Thread displayName through subcommand registrars for 'Run <cli> ...' hints
## Background

#83 / #89 introduced a `displayName` (separate from identity `cliName`) so consumer CLIs delegating to the shared `apijack` binary can brand themselves in `--help`, the setup banner, and the auth-failure `Run '<cli> setup'` hints. The split was kept tight to the locations called out in #83's acceptance criteria.

The first-pass review on #89 surfaced that several subcommand-registrar error/help hints still interpolate `cliName` instead of `displayName`. Same root cause as #83 — secondary error paths the original issue didn't explicitly enumerate. A consumer that sets `programName: "mycli"` will still see `Run 'apijack ...'` strings when they hit any of these code paths, which is confusing.

## Locations

- `src/commands/config/register.ts:33` — `No environments configured. Run '${cliName} setup' to add one.`
- `src/commands/config/register.ts:146` — `No environments configured. Run '${cliName} config import' first.`
- `src/commands/routine/register.ts:73` — `Run '${cliName} routine init' to install built-in routines.`
- `src/commands/routine/register.ts:104` — `No active session. Run '${cliName} setup' first.`
- `src/commands/routine/register.ts:110` — `No active session. Run '${cliName} setup' first.`
- `src/commands/routine/register.ts:233` — `No active session. Run '${cliName} setup' first.`
- `src/commands/config/update-password/update-password.ts:21` — `No environments configured. Run '${deps.cliName} config import' first.`

## Out of scope (intentionally keeps `cliName`)

- `~/.${cliName}/config.json` path-log (real on-disk path, not a hint).
- MCP routines dir construction (`join(homedir(), '.' + options.name, 'routines')`).
- `src/run-routine.ts` — programmatic callers using core directly; "Run 'apijack setup'" is the truthful instruction in that mode.
- Anywhere `cliName` keys storage paths, env-var prefixes, or session keys.

## Implementation note

The registrars receive their `cliName` via deps/parameters today. Either:
1. Thread `displayName` through the same channel, defaulting to `cliName` when omitted (matches the pattern in `cli-builder.ts` and `setup.ts`).
2. Add `displayName` to the dep struct alongside `cliName`.

Option 2 is probably less invasive — most call sites already accept a deps object.

## Acceptance criteria

- [ ] All seven listed sites use `displayName` (with `cliName` fallback) instead of `cliName`.
- [ ] A consumer that sets `programName: "mycli"` sees `Run 'mycli setup'`, `Run 'mycli config import'`, etc., when hitting these error paths.
- [ ] Storage paths, env-var prefix, and session keys still use the identity `cliName` (no path-log lies).
- [ ] Unit tests cover at least one site per file showing the displayName flow with and without `programName` set.

garretpremo:
Closes #91

## Summary

Follow-up to #83/#89. Seven user-facing `Run '<cli> ...'` hints across the config and routine subcommand registrars still interpolated the identity `cliName` instead of the new `displayName`. A consumer that sets `programName: "mycli"` would see `Run 'apijack setup'` (or similar) when hitting any of these error paths, breaking the branding contract that #83 established.

This PR plumbs `displayName` through both registrars (opts struct on `registerConfigCommand`, optional 8th positional on `registerRoutineCommand`) and the `configUpdatePasswordAction` deps, defaulting to `cliName` when omitted so existing consumers see no behavior change. Identity-keyed uses (`~/.${cliName}/` path log, MCP routines dir, `run-routine.ts`, storage/env-var/session keys) intentionally keep `cliName` per the issue's explicit out-of-scope list.

### Plumbing

- `src/commands/config/update-password/update-password.ts` — `ConfigUpdatePasswordDeps` gains `displayName?: string`; the `No environments configured` throw uses `deps.displayName ?? deps.cliName`.
- `src/commands/config/register.ts` — `registerConfigCommand` opts gain `displayName?`. Computed once at the top: `const displayName = opts?.displayName ?? cliName`. Used at the two hint sites (`config list` empty + `update-password` empty-envs guard) and forwarded into `configUpdatePasswordAction` deps.
- `src/commands/routine/register.ts` — `registerRoutineCommand` gains an optional 8th positional param `displayName?`. Local `const display = displayName ?? cliName` (named `display` to avoid shadowing the parameter). Used at the four hint sites: `routine list` empty branch, `routine run` no-dispatch JSON failure, `routine run` no-dispatch stderr, `routine test` no-dispatch.
- `src/cli-builder.ts` — passes `displayName` (already computed at line 139) to both registrars.

### Tests

`tests/cli-builder.test.ts` gains a new `describe('displayName threaded through subcommand registrar hints (#91)')` block with 6 tests:

- `config list`: hint shows `programName` when set; falls back to `cliName` when omitted.
- `routine list`: hint shows `programName` when set; falls back to `cliName` when omitted.
- `configUpdatePasswordAction` (direct): throw includes `displayName` when set; falls back to `cliName` when omitted.

## Acceptance criteria

- [x] All seven listed sites use `displayName` (with `cliName` fallback) instead of `cliName`
- [x] A consumer that sets `programName: "mycli"` sees `Run 'mycli setup'`, `Run 'mycli config import'`, etc., on these error paths
- [x] Storage paths, env-var prefix, and session keys still use the identity `cliName` (no path-log lies)
- [x] Unit tests cover at least one site per file showing the displayName flow with and without `programName` set

## Test plan

- [x] `bun test` — 874 pass / 0 fail (was 868; +6 new tests)
- [x] `bun run lint` — 0 errors (108 pre-existing warnings unchanged)
- [x] `grep -rn "Run '\${cliName}\|Run '\${deps\.cliName}" src/` — zero remaining hint sites
- [x] Self-review via `superpowers:code-reviewer` — verdict: ready to merge

### DIFF
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -451,6 +451,7 @@ export function createCli(options: CliOptions): Cli {
                 configPath: options.configPath,
                 knownSites: options.knownSites,
                 allowedCidrs: options.allowedCidrs,
+                displayName,
             });
             registerUpgradeCommand(program, options.version);
             registerMcpCommand(
@@ -820,7 +821,7 @@ export function createCli(options: CliOptions): Cli {
             }
 
             // 12. Register routine commands
-            registerRoutineCommand(program, cliName, routinesDir, dispatch, options.builtinRoutinesDir, customResolvers, pluginRegistry);
+            registerRoutineCommand(program, cliName, routinesDir, dispatch, options.builtinRoutinesDir, customResolvers, pluginRegistry, displayName);
 
             // 13. Handle -o routine-step
             if (isRoutineStep) {
--- a/src/commands/config/register.ts
+++ b/src/commands/config/register.ts
@@ -14,9 +14,12 @@ export function registerConfigCommand(
         configPath?: string;
         knownSites?: Record<string, { url: string; description: string; group?: string }>;
         allowedCidrs?: string[];
+        /** Display name for user-facing hints. Defaults to cliName. */
+        displayName?: string;
     },
 ): void {
     const configOpts = opts?.configPath ? { configPath: opts.configPath } : undefined;
+    const displayName = opts?.displayName ?? cliName;
     const config = program
         .command('config')
         .description('Manage environment configurations');
@@ -30,7 +33,7 @@ export function registerConfigCommand(
             });
 
             if (envs.length === 0) {
-                console.log(`No environments configured. Run '${cliName} setup' to add one.`);
+                console.log(`No environments configured. Run '${displayName} setup' to add one.`);
 
                 return;
             }
@@ -143,7 +146,7 @@ export function registerConfigCommand(
                 const cfg = await loadConfig(cliName, configOpts);
 
                 if (!cfg || Object.keys(cfg.environments).length === 0) {
-                    console.error(`No environments configured. Run '${cliName} config import' first.`);
+                    console.error(`No environments configured. Run '${displayName} config import' first.`);
                     process.exit(1);
                 }
 
@@ -170,6 +173,7 @@ export function registerConfigCommand(
                         loadConfig: async () => cfg,
                         save: saveEnvironment,
                         cliName,
+                        displayName,
                         saveOpts: configOpts ?? {},
                     });
                     console.log('Password updated.');
--- a/src/commands/config/update-password/update-password.ts
+++ b/src/commands/config/update-password/update-password.ts
@@ -6,6 +6,8 @@ export interface ConfigUpdatePasswordDeps {
     loadConfig: (cliName: string) => Promise<CliConfig | null>;
     save: (cliName: string, name: string, env: EnvironmentConfig, setActive?: boolean, opts?: Record<string, unknown>) => Promise<void>;
     cliName: string;
+    /** Display name for user-facing hints. Defaults to cliName. */
+    displayName?: string;
     saveOpts: Record<string, unknown>;
 }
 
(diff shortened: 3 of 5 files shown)


garretpremo (commented):

PR cleanly addresses the seven hint sites called out in #91. All acceptance criteria met: every site uses `displayName` with `cliName` fallback, identity-keyed paths/env vars/session keys remain on `cliName`, and tests cover the displayName flow with and without `programName` for each of the three modified source files. CI green, lint clean, 874/874 tests pass.

<details>

- `registerRoutineCommand` now takes 8 positional params. Adding `displayName` as a positional matches the existing pattern, but the signature is getting long — a future refactor to an opts object would be reasonable, especially if more cross-cutting fields like this surface.
- Slight naming asymmetry: `commands/config/register.ts` uses `displayName` as the local var, `commands/routine/register.ts` uses `display` to avoid shadowing the parameter. Justified in the PR body — flagging only because a reader skimming both files side by side may pause on it.

</details>

<details>
<summary>Nitpicks</summary>

- `routine/register.ts` has four hint sites; only the `routine list` site has direct test coverage. AC requires "at least one site per file" so this is met, but one extra test for the `routine run` no-dispatch path (e.g., `routine run foo` without an active session) would lock in the other three sites against future regression.

</details>

garretpremo (commented):

Approved.

- `registerRoutineCommand` now takes 8 positional params. Adding `displayName` as a positional matches the existing pattern, but the signature is getting long — a future refactor to an opts object would be reasonable, especially if more cross-cutting fields like this surface.
- Slight naming asymmetry: `commands/config/register.ts` uses `displayName` as the local var, `commands/routine/register.ts` uses `display` to avoid shadowing the parameter. Justified in the PR body — flagging only because a reader skimming both files side by side may pause on it.

PR #102, opened 2026-05-05
fix: preserve original 401/403 when refresh callback throws (#98)

### LINKED ISSUE #77: SessionAuthStrategy: auto-refresh and retry once on stale-session 401/403

### LINKED ISSUE #98: SessionAuthStrategy: preserve original 401/403 (with refresh error as `cause`) when refresh callback throws
Spawned from #94 (closes #77).

## Background

The stale-session refresh + retry mechanism in `src/codegen/client.ts:229-247` looks like:

```js
if (!res.ok && this.onRefreshNeeded && this.refreshOn?.includes(res.status)) {
  await this.onRefreshNeeded();   // ← refresh-callback failure path
  // ... retry once
}
```

If `onRefreshNeeded()` throws (e.g., the credential rotation refresh itself fails because creds are now invalid, or `/session` returns 5xx), the refresh error propagates as-is and the original `{status, body}` from the 401/403 response is lost. The caller never sees that the underlying request was a 401 — they see a refresh-related error instead.

## What #77 originally asked for

> If the refresh itself fails, propagate the original error untouched.

This was deferred during #77's triage to keep that PR scoped to test + docs. The agreed-on approach (when implemented) is to throw the **original** `{status, body}` so existing catchers continue to work, and attach the refresh error as `cause` per the standard `Error` API so debugging info isn't lost. Roughly:

```js
const originalBody = await res.text();
try {
  await this.onRefreshNeeded();
} catch (refreshErr) {
  throw Object.assign(new Error(`Request failed with ${res.status}`), {
    status: res.status,
    body: originalBody,
    cause: refreshErr,
  });
}
// retry...
```

## Acceptance criteria

- [ ] When `onRefreshNeeded()` throws, the thrown error carries the original `{status, body}` shape (so existing catchers that check `err.status === 401` still work).
- [ ] The refresh failure is preserved as `err.cause` so debugging info isn't lost.
- [ ] An integration test covers this path (extending `tests/auth/session-auth-stale-retry.integration.test.ts` is fine).
- [ ] Update the in-test breadcrumb at `tests/auth/session-auth-stale-retry.integration.test.ts:124-127` once covered.

## Notes

- Touches `src/codegen/client.ts` (codegen — the generated `ApiClient.request()`).
- Same retry path is reused for any future `refreshOn` consumer, so the behavior change benefits anyone who opts into the mechanism — not just `SessionAuthStrategy`.
- Ship behind no flag. The current behavior (refresh error wins) is observably worse than the proposed behavior in every case.

garretpremo:
Closes #98

## Summary

When `SessionAuthStrategy` (or any consumer wired into the generated `ApiClient`'s `refreshOn` mechanism) gets a 401/403 and the refresh callback throws, the original status/body was lost — callers checking `err.status === 401` saw a refresh-related error instead. This change preserves the original error shape and attaches the refresh failure as `cause` per the standard `Error` API.

Spawned from #94 (which closed #77 with the test + docs scope; this completes the originally-asked-for behavior change deferred during that triage).

### `src/codegen/client.ts`

In the generated `ApiClient.request()`, the refresh-failure branch now:
1. Reads the original response body **before** invoking the refresh callback (so it isn't lost when the callback throws).
2. Wraps the refresh call in `try`/`catch`.
3. On failure, throws an `Error` carrying the original `{status, body}` and the refresh failure as `cause`.

```js
if (!res.ok && this.onRefreshNeeded && this.refreshOn?.includes(res.status)) {
  const originalBody = await res.text();
  try {
    await this.onRefreshNeeded();
  } catch (refreshErr) {
    throw Object.assign(new Error(`Request failed with ${res.status}`), {
      status: res.status,
      body: originalBody,
      cause: refreshErr,
    });
  }
  // ... retry once (unchanged)
}
```

Scope is intentionally limited to the refresh-failure branch — the other throw sites at the surrounding lines (retry-failed and non-refresh-failed) keep their existing shape, matching the issue's narrow scope.

### `tests/auth/session-auth-stale-retry.integration.test.ts`

- Removed the deferred-follow-up breadcrumb (lines 124–127 in the prior file).
- Added a new integration test that:
  - Makes a request that returns 401.
  - Wires a refresh callback that throws.
  - Asserts the thrown value is an `Error` with `status === 401`, the original response body, and the refresh error as `cause`.
  - Asserts no retry was attempted (only one DELETE call).

## Acceptance criteria

- [x] When `onRefreshNeeded()` throws, the thrown error carries the original `{status, body}` shape (existing `err.status === 401` catchers still work).
- [x] The refresh failure is preserved as `err.cause`.
- [x] Integration test covers the new behavior.
- [x] Breadcrumb at `tests/auth/session-auth-stale-retry.integration.test.ts:124-127` updated.

## Test plan

- [x] `bun test` — 875 pass / 0 fail
- [x] `bun run lint` — 0 errors (108 pre-existing warnings)
- [x] New test verifies the error shape end-to-end through the generated `ApiClient`.


garretpremo (commented):

LGTM. Implementation matches the issue's prescribed approach: the original response body is read before the refresh callback runs, the callback is wrapped in `try`/`catch`, and on failure an `Error` carrying `{status, body, cause}` is thrown so existing `err.status === 401` catchers keep working. Integration test exercises the full path through the generated `ApiClient` and asserts `instanceof Error`, status, body, cause, and no retry. CI is green across ubuntu + windows.

<details>

- Throw-shape inconsistency at the three sites in `ApiClient.request()`: the new refresh-failure branch throws `Error` (with augmented `status`/`body`/`cause`), while the retry-failed branch (`src/codegen/client.ts:242`) and the non-refresh failure branch (`src/codegen/client.ts:250`) still throw plain objects `{ status, body }`. The PR explicitly scopes this fix narrowly per the issue, so callers that only check `err.status` are unaffected — but `instanceof Error` discriminators will now diverge across the three failure modes. Worth a follow-up to unify on the `Error`-with-properties shape if/when there's a reason to touch those other sites.

</details>

garretpremo (commented):

Approved.

- Throw-shape inconsistency at the three sites in `ApiClient.request()`: the new refresh-failure branch throws `Error` (with augmented `status`/`body`/`cause`), while the retry-failed branch (`src/codegen/client.ts:242`) and the non-refresh failure branch (`src/codegen/client.ts:250`) still throw plain objects `{ status, body }`. The PR explicitly scopes this fix narrowly per the issue, so callers that only check `err.status` are unaffected — but `instanceof Error` discriminators will now diverge across the three failure modes. Worth a follow-up to unify on the `Error`-with-properties shape if/when there's a reason to touch those other sites.

PR #85, opened 2026-05-01
Programmatic runRoutine() API + --json CLI mode

garretpremo:
Closes #84

## Summary

Adds a programmatic `runRoutine()` API and `apijack routine run <name> --json` mode so consumers (Playwright fixtures, Vitest setups, Bun scripts) can invoke routines as data, not just as CLI side-effects. Two surfaces, one engine, one bootstrap path: `cli.runRoutine(...)` for projects with a `bin/<cli>.ts`, and standalone `runRoutine(...)` for projects on the shared `apijack` binary.

### Engine (`src/routine/executor.ts`)

`RoutineResult` is extended with `status: 'ok' | 'failed'`, `output: Record<string, unknown>` (named outputs from `output:` aliases), `steps: RoutineResultStep[]` (per-step entries with status), and `durationMs`. `success`/`stepsRun`/`stepsSkipped`/`stepsFailed` are preserved for back-compat. New `ExecutorOptions.silent` gates the executor's stderr/console writes (default `false` keeps existing CLI behavior identical). Failed dispatches and assertion failures correctly do **not** populate `output[alias]`.

### Programmatic API (`src/cli-builder.ts`, `src/run-routine.ts`)

- `cli.runRoutine(name, opts?)` — method on the `Cli` returned by `createCli`. Calls a new private idempotent `_buildRoutineRuntime()` that resolves auth, builds the dispatcher, and caches the result.
- `runRoutine(name, opts?)` — top-level export from `@apijack/core`. Mirrors `bin/apijack.ts`'s bootstrap (env loading, project extension wiring, auth strategy resolution) and delegates.
- Failure semantics: programmatic resolves with `{ status: 'failed', steps, output: <partial> }` on step failure; throws only on engine errors (routine not found, validation errors, no active env config).

### CLI (`src/commands/routine/register.ts`, `run/run.ts`)

`apijack routine run <name> --json` emits `JSON.stringify(result)` to stdout, exits non-zero on failure, suppresses progress output. Default mode is unchanged. Synthesized failure shapes are typed as `RoutineResult` so future drift is caught at compile time.

### Public surface (`src/index.ts`)

Adds `runRoutine`, `RoutineResult`, `RoutineResultStep`, `RunRoutineOptions`, `StandaloneRunRoutineOptions`.

### Docs (`CLAUDE.md`)

New "Programmatic Routine Invocation" section + `--json` line in the Routine Commands list.

## Acceptance criteria

[...]

PR #104, opened 2026-05-05
fix: emit steps[] entry for forEach non-array path and tag iterations with index (#100)

### LINKED ISSUE #100: RoutineResult: emit `steps[]` entry for `forEach` non-array path; tag iterations with index
## Background

PR #84 / #85 (v1.11.0) introduced the programmatic `runRoutine()` API plus the `--json` CLI mode, both of which expose a structured `RoutineResult` (with `steps[]`, `stepsRun`, `stepsSkipped`, `stepsFailed`) for downstream consumers. The bot review on #85 flagged two `forEach` reporting issues that were never tracked separately. Both still reproduce.

## Issue 1: non-array `forEach` skip emits no `steps[]` entry

`src/routine/executor.ts:178-188`:

```ts
if (!Array.isArray(rawItems)) {
    if (!this.options?.silent) {
        process.stderr.write(`Warning: forEach on "${step.name}" did not resolve to an array\n`);
    }
    this.stepsSkipped++;
    return true;
}
```

Compare with the condition-skip path (`executor.ts:122-124`), which does both:

```ts
this.stepsSkipped++;
this.steps.push({ name: step.name, status: 'skipped' });
```

The asymmetry means a programmatic consumer reading `result.stepsSkipped` sees N, but `result.steps` only has N - <forEach-non-array-count> entries — they can't tell which step skipped without parsing the stderr warning.

## Issue 2: forEach iteration entries are indistinguishable

`src/routine/executor.ts:209-230` runs `runSteps(step.steps!, iterCtx)` once per item. Each inner step's `executeCommand` pushes a `steps[]` entry tagged with the inner step's `name`. Across N iterations, you get N entries with the same name and no iteration index. A consumer iterating `result.steps` to pair input with output has no clean way to say "this output came from iteration 3".

## Proposed fix

For Issue 1: emit `{ name: step.name, status: 'skipped' }` (matching the condition-skip path) before returning. One line.

For Issue 2: thread the iteration index into the pushed entry. Options:
- Add an `iteration?: { index: number; total: number }` field on `RoutineStepResult`.
- Suffix the `name`: `name: \`${innerStep.name}[${i}]\``.

The first option is structurally cleaner; the second is one-line. Either resolves the consumer pain.

## Acceptance criteria

- [ ] `result.stepsSkipped === result.steps.filter(s => s.status === 'skipped').length` always holds (closes Issue 1).
- [ ] When a `forEach` block produces N iterations, the consumer can identify which `steps[]` entry came from which iteration (closes Issue 2).
- [ ] Unit test for the non-array skip path: `forEach: '$not-an-array'` → asserts `result.stepsSkipped === 1` AND a corresponding `steps[]` entry exists with `status: 'skipped'`.
- [ ] Unit test for iteration tagging: `forEach: ['a', 'b', 'c']` with one inner step → asserts the produced `steps[]` entries are distinguishable by iteration.

## Notes

- Public-API impact: adding a new field to `RoutineStepResult` is additive (no break). Suffixing the `name` is technically a string-shape change that downstream consumers might pattern-match — pick the structural-field option if there's any concern.
- Both fixes ship together — they're the same defect category in the same function.
- Same retry/iteration plumbing is reused by `range:` (`runRange`), so the iteration-index fix benefits range steps too.

garretpremo:
Closes #100

## Summary

Fixes two related defects in `RoutineResult` reporting around `forEach` (and by extension `range:`) — both surfaced by the bot review on #85 but never tracked separately. Both made `result.steps[]` an unreliable mirror of what actually executed for any consumer of the programmatic `runRoutine()` API or the `--json` CLI mode introduced in v1.11.0.

1. **Non-array `forEach` skip emitted no `steps[]` entry** — `stepsSkipped` was incremented but the array was not, breaking the invariant that the condition-skip path maintains.
2. **`forEach` iteration entries were indistinguishable** — N inner-step entries shared the same `name` with no marker linking them to the iteration that produced them.

### `src/routine/executor.ts`

- **New optional field on `RoutineResultStep`**: `iteration?: { index: number; total: number }`. Additive — backward compatible.
- **New private `pushStep()` helper**: stamps `iteration` automatically when called from inside an iteration. All four `this.steps.push(...)` sites now go through it.
- **Iteration tracking via `currentIteration`**: set per iteration in `runIteration` (which is the shared engine for both `forEach` and `range:`). Save/restore around the loop so nested iterations work too.
- **Non-array `forEach` path** now emits `{ name: step.name, status: 'skipped' }` — mirroring the condition-skip path two-line pattern.
- **Assert-fail overwrite path** (`executor.ts:303-307`) preserves the iteration tag when rewriting the last entry from `ok` to `failed`.

### `tests/routine/executor.test.ts`

Seven new tests:

- non-array `forEach` produces a skipped `steps[]` entry and the `stepsSkipped == steps.filter(skipped).length` invariant holds
- `forEach` iteration entries tagged with `{ index, total }`
- `range:` iteration entries tagged the same way (shared engine)
- skipped step inside a `forEach` iteration carries the iteration tag
- assertion failure inside a `forEach` iteration preserves the iteration tag
- non-iteration steps do not carry an iteration tag (no leakage)
- steps after a `forEach` block do not carry an iteration tag (save/restore works)

Iteration `index` is 0-based to match standard array indexing in structured data; `total` is the array length. The user-facing `onIteration` callback continues to pass `i + 1` (1-based) for terminal display — that surface is unchanged.

## Acceptance criteria

- [x] `result.stepsSkipped === result.steps.filter(s => s.status === 'skipped').length` always holds
- [x] For an N-iteration `forEach`, the consumer can identify which `steps[]` entry came from which iteration (via `step.iteration.index`)
- [x] Unit test: `forEach: '$not-an-array'` → `stepsSkipped === 1` AND a `steps[]` entry with `status: 'skipped'` exists
- [x] Unit test: `forEach: ['a','b','c']` with one inner step → produced entries are distinguishable by iteration

## Test plan

- [x] `bun test` — 881 pass / 0 fail
- [x] `bun run lint` — 0 errors (108 pre-existing warnings)

### DIFF
--- a/src/routine/executor.ts
+++ b/src/routine/executor.ts
@@ -21,6 +21,7 @@ export interface RoutineResultStep {
     status: 'ok' | 'failed' | 'skipped';
     output?: unknown;
     error?: string;
+    iteration?: { index: number; total: number };
 }
 
 export interface RoutineResult {
@@ -55,8 +56,17 @@ class RoutineExecutor {
     private stepsSkipped = 0;
     private stepsFailed = 0;
     private inIteration = false;
+    private currentIteration: { index: number; total: number } | null = null;
     private steps: RoutineResultStep[] = [];
     private namedOutput: Record<string, unknown> = {};
+
+    private pushStep(entry: RoutineResultStep): void {
+        if (this.currentIteration) {
+            entry.iteration = { ...this.currentIteration };
+        }
+
+        this.steps.push(entry);
+    }
 
     constructor(
         private dispatch: CommandDispatcher,
@@ -121,7 +131,7 @@ class RoutineExecutor {
 
             if (!evaluateCondition(step.condition, ctx)) {
                 this.stepsSkipped++;
-                this.steps.push({ name: step.name, status: 'skipped' });
+                this.pushStep({ name: step.name, status: 'skipped' });
 
                 if (this.options?.onStep && !this.inIteration)
                     this.options.onStep(step, i, steps.length);
@@ -183,6 +193,7 @@ class RoutineExecutor {
             }
 
             this.stepsSkipped++;
+            this.pushStep({ name: step.name, status: 'skipped' });
 
             return true;
         }
@@ -205,6 +216,7 @@ class RoutineExecutor {
 
         const asName = step.as || 'item';
         this.inIteration = true;
+        const previousIteration = this.currentIteration;
 
         for (let i = 0; i < items.length; i++) {
             if (this.options?.onIteration)
@@ -216,6 +228,8 @@ class RoutineExecutor {
                     totalSteps,
                 );
 
+            this.currentIteration = { index: i, total: items.length };
+
             const iterCtx: RoutineContext = {
                 ...ctx,
                 forEachItem: { name: asName, value: items[i] },
@@ -224,12 +238,14 @@ class RoutineExecutor {
 
             if (!ok && !step.continueOnError) {
                 this.inIteration = false;
+                this.currentIteration = previousIteration;
 
                 return false;
             }
         }
 
         this.inIteration = false;
+        this.currentIteration = previousIteration;
 
         if (this.options?.onIteration && !this.options.silent) process.stderr.write('\n');
 
@@ -284,7 +300,7 @@ class RoutineExecutor {
             if (step.output) ctx.stepOutputs.set(step.output, stepResult);
 
             this.stepsRun++;
-            this.steps.push({ name: step.name, status: 'ok', output: result });
+            this.pushStep({ name: step.name, status: 'ok', output: result });
 
             if (step.output) this.namedOutput[step.output] = result;
 
@@ -299,12 +315,19 @@ class RoutineExecutor {
                     stepResult.success = false;
                     stepResult.error = `Assertion failed: ${step.assert}`;
                     this.stepsFailed++;
-                    // overwrite the previously-pushed 'ok' with the failed entry
-                    this.steps[this.steps.length - 1] = {
+                    // overwrite the previously-pushed 'ok' with the failed entry,
+                    // preserving the iteration tag if any
+                    const overwrite: RoutineResultStep = {
                         name: step.name,
                         status: 'failed',
                         error: stepResult.error,
                     };
+
+                    if (this.currentIteration) {
+                        overwrite.iteration = { ...this.currentIteration };
+                    }
+
+                    this.steps[this.steps.length - 1] = overwrite;
 
                     if (step.output) delete this.namedOutput[step.output];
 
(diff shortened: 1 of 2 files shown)


garretpremo (commented):

LGTM. Both defects from #100 are addressed with the structural-field option (correctly preferred per the issue's notes). All four explicit acceptance criteria are met, the seven new tests are well-targeted (skip path, forEach + range tagging, nested condition-skip, assert-fail rewrite preserving the tag, no leakage before/after the loop), and CI is green. Additive `iteration?` field on `RoutineResultStep` keeps the `@apijack/core` surface backward compatible.

<details>

- `currentIteration` is correctly save/restored across (potentially nested) `runIteration` calls, but the sibling `inIteration` boolean (`executor.ts:207`, `executor.ts:226`, `executor.ts:232`) still uses simple set/clear. In a nested-forEach scenario, the inner loop's terminal `this.inIteration = false` will leak into the outer loop's remaining iterations, and the only consumer (`if (this.options?.onStep && !this.inIteration)` at `executor.ts:245`) would start firing per-step terminal callbacks while the outer iteration is still running. Pre-existing latent issue — not introduced by this PR — but the PR is in this exact area, so flagging it. Could be unified by deriving `inIteration` from `this.currentIteration !== null` (or by giving `inIteration` the same save/restore treatment) in a follow-up.
- The assertion-fail overwrite at `executor.ts:315-330` re-implements the iteration-stamping inline instead of going through `pushStep`. Reasonable since it's an in-place rewrite rather than a push, but a small `pushStep`-like helper that takes an index (or a `replaceLastStep`) could deduplicate the `if (this.currentIteration) entry.iteration = { ...this.currentIteration }` logic. Minor.

</details>

<details>
<summary>Nitpicks</summary>

- PR title uses `fix:` and the changelog will treat this as patch. The change is genuinely a defect fix per the issue framing, so this is fine — but it does add a new optional field to a public type, which is technically additive surface area. If a future reader scans the patch-bump release notes for new fields, they won't find this one. No action needed; just noting.

</details>

garretpremo (commented):

Approved.

- `currentIteration` is correctly save/restored across (potentially nested) `runIteration` calls, but the sibling `inIteration` boolean (`executor.ts:207`, `executor.ts:226`, `executor.ts:232`) still uses simple set/clear. In a nested-forEach scenario, the inner loop's terminal `this.inIteration = false` will leak into the outer loop's remaining iterations, and the only consumer (`if (this.options?.onStep && !this.inIteration)` at `executor.ts:245`) would start firing per-step terminal callbacks while the outer iteration is still running. Pre-existing latent issue — not introduced by this PR — but the PR is in this exact area, so flagging it. Could be unified by deriving `inIteration` from `this.currentIteration !== null` (or by giving `inIteration` the same save/restore treatment) in a follow-up.
- The assertion-fail overwrite at `executor.ts:315-330` re-implements the iteration-stamping inline instead of going through `pushStep`. Reasonable since it's an in-place rewrite rather than a push, but a small `pushStep`-like helper that takes an index (or a `replaceLastStep`) could deduplicate the `if (this.currentIteration) entry.iteration = { ...this.currentIteration }` logic. Minor.

PR #107, opened 2026-05-06
feat: add dropBaseHeaders flag to SessionAuthConfig (#106)

### LINKED ISSUE #106: SessionAuthStrategy: base-strategy headers (e.g. Authorization) persist on every post-handshake request
## Summary

After `SessionAuthStrategy` completes the `/session` handshake, the `AuthSession` it returns includes the headers from the wrapped base strategy unchanged (`headers: baseSession.headers` in `auth/session-auth.ts`). For a `BasicAuthStrategy` base, that means `Authorization: Basic <…>` is then mirrored onto every subsequent API request through `resolveRequestHeaders`. There is no documented way to drop those base headers after the handshake.

For Spring Security backends with stateful 2FA, this re-triggers the 2FA filter on every API call: the server treats each request as a fresh authentication attempt rather than a session-cookie-authenticated one, and either re-prompts, 401s, or sets `Set-Cookie: Max-Age=0` and invalidates the active session.

## Repro (via `@apijack/core` against a Spring Security backend with 2FA enabled)

1. Configure an env with a `sessionAuth` block (`session.endpoint = "/session"`, `cookies.extract = ["SESSION", "XSRF-TOKEN"]`, an `onChallenge` handler) and `BasicAuthStrategy` as the base.
2. Run any command. The handshake completes and 2FA is satisfied; the cached `AuthSession` includes both `Authorization: Basic …` (from the base) and the captured cookies.
3. Inspect the request headers on the next API call: both `Authorization: Basic …` and `Cookie: SESSION=…` are sent.
4. Server response varies by configuration: re-prompts for 2FA, 401s, or invalidates the session.

## Expected

`SessionAuthStrategy` should provide a documented way to drop base-strategy headers from the returned `AuthSession`, so post-handshake requests carry only what the consumer explicitly wants (typically just `Cookie` plus any `headerMirror` headers).

Most natural shape: a flag on `SessionAuthConfig` (e.g. `dropBaseHeaders: true`), or a list of header names to omit. Strategies that legitimately want to keep base headers (e.g. layering Bearer + session) can opt out.

## Notes

- Today the only workaround is to bypass `SessionAuthStrategy` entirely and write a custom `AuthStrategy` that does the `/session` + `onChallenge` handshake itself, then returns an `AuthSession` containing only cookie headers. That works but means the consumer also gives up the niceties added in #77 (`refreshOn`) and `headerMirror` support without re-implementing them.
- A consumer that rolls its own strategy can populate `session.cookies` to make `headerMirror` fire — but only if `sessionAuth` is also set in env config, which then re-wraps the custom strategy in `SessionAuthStrategy`. There's currently no clean way to opt into `headerMirror` without the wrap.

## Acceptance criteria

- `SessionAuthConfig` exposes a documented mechanism for dropping (or filtering) base-strategy headers after handshake. `dropBaseHeaders: true` is the simplest shape.
- Integration test: with `BasicAuthStrategy` base + the new flag, post-handshake `AuthSession.headers` does NOT include `Authorization`.
- Default behavior unchanged for backwards compatibility.
- Brief note in auth strategy docs covering when to enable.

garretpremo:
Closes #106

## Summary

Adds an opt-in `dropBaseHeaders` flag to `SessionAuthConfig`. When enabled, `SessionAuthStrategy` strips the wrapped base strategy's headers (e.g. `Authorization: Basic …`) from the returned `AuthSession` after the `/session` handshake completes. Base headers are still sent to the session endpoint itself; only post-handshake API calls carry just cookies + `headerMirror` headers.

The motivating case is stateful backends like Spring Security with 2FA: re-presenting the base credentials on every API call re-triggers the auth filter and either re-prompts for 2FA, 401s, or invalidates the active session. Today the only workaround is to write a custom `AuthStrategy` from scratch — which sacrifices `refreshOn`, `headerMirror`, and `onChallenge` support.

Default behavior is unchanged.

### Public API

- `SessionAuthConfig.dropBaseHeaders?: boolean` — new optional flag (`src/auth/types.ts`).

### Behavior

`SessionAuthStrategy` (`src/auth/session-auth.ts`):

- `authenticate()` returns `headers: this.config.dropBaseHeaders ? {} : baseSession.headers` after a successful handshake. The `fetch(/session)` call itself still uses `baseSession.headers`, so handshake auth (Basic, etc.) works unchanged.
- `restore()` mirrors the same logic — otherwise a process restart with a cached session on disk would re-introduce `Authorization` and undo the fix.
- `onChallenge` retry path is unaffected: retries inside `authenticate()` still build `retryHeaders` from `baseSession.headers`, so 2FA flows continue to work; only the final returned session is filtered.

```yaml
# example consumer config
new SessionAuthStrategy(new BasicAuthStrategy(), {
  session: { endpoint: '/session' },
  cookies: { extract: ['SESSION'], applyTo: ['POST', 'PUT', 'DELETE'] },
  dropBaseHeaders: true,
});
```

### Docs

`CLAUDE.md` — new "Dropping base-strategy headers post-handshake (opt-in)" subsection under Authentication Strategies, with a copy-pasteable example and an explicit note that the handshake itself still sends base headers.

### Tests

`tests/auth/session-auth.test.ts` — new `describe('dropBaseHeaders', …)` block with 7 tests covering:

- Returned session omits base headers when enabled
- `/session` handshake request still sends base headers when enabled
- Cookies preserved when base headers are dropped
- `restore()` omits base headers from returned session when enabled
- Default config (flag absent) keeps base headers — backwards-compat pin
- `onChallenge` retry sends base headers but returned session drops them — pins the 2FA path
- Downstream `resolveRequestHeaders` produces `Cookie` + `headerMirror` only — end-to-end pin

## Acceptance criteria

- [x] `SessionAuthConfig` exposes a documented mechanism for dropping base-strategy headers after handshake (`dropBaseHeaders: true`).
- [x] Integration test: with `BasicAuthStrategy` base + the new flag, post-handshake `AuthSession.headers` does NOT include `Authorization`.
- [x] Default behavior unchanged for backwards compatibility.
- [x] Brief note in auth strategy docs covering when to enable.

## Test plan

- [x] `bun test` — 892 pass / 0 fail
- [x] `bun run lint` — 0 errors (109 pre-existing warnings)
- [x] Self-review by code-reviewer subagent — Important issue raised (untested `onChallenge` retry path under flag) addressed; Minor TSDoc clarification applied.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -216,6 +216,20 @@ new SessionAuthStrategy(new BasicAuthStrategy(), {
 
 When the generated client receives a status in `refreshOn`, it calls the wired refresh callback — which invalidates the cached session and re-bootstraps `/session` — then retries the original request once. Capped at one retry. Strategies that don't use `/session` are unaffected (the field is ignored).
 
+### Dropping base-strategy headers post-handshake (opt-in)
+
+By default, `SessionAuthStrategy` mirrors the wrapped base strategy's headers (e.g. `Authorization: Basic …` from `BasicAuthStrategy`) onto every post-handshake API request. For stateful backends — Spring Security with 2FA, for instance — re-presenting the base credentials on every call re-triggers the auth filter and either re-prompts, 401s, or invalidates the active session. Opt in to `dropBaseHeaders` to strip them:
+
+```ts
+new SessionAuthStrategy(new BasicAuthStrategy(), {
+  session: { endpoint: '/session' },
+  cookies: { extract: ['SESSION'], applyTo: ['POST', 'PUT', 'DELETE'] },
+  dropBaseHeaders: true,
+});
+```
+
+Base headers are still sent to the `/session` endpoint itself (handshake works as before); only post-handshake API calls carry just cookies + any `headerMirror` headers. Default is `false` for backwards compatibility.
+
 ## Project Extensions
 
 The `.apijack/` directory at a project root is auto-loaded when the CLI runs inside that project:
--- a/src/auth/session-auth.ts
+++ b/src/auth/session-auth.ts
@@ -69,7 +69,7 @@ export class SessionAuthStrategy implements AuthStrategy {
         const cookies = { ...challengeCookies, ...this.extractCookies(res) };
 
         return {
-            headers: baseSession.headers,
+            headers: this.config.dropBaseHeaders ? {} : baseSession.headers,
             cookies,
             expiresAt: baseSession.expiresAt,
             data: baseSession.data,
@@ -86,7 +86,7 @@ export class SessionAuthStrategy implements AuthStrategy {
         if (!baseRestored) return null;
 
         return {
-            headers: baseRestored.headers,
+            headers: this.config.dropBaseHeaders ? {} : baseRestored.headers,
             cookies: cached.cookies,
             expiresAt: baseRestored.expiresAt,
             data: baseRestored.data,
--- a/src/auth/types.ts
+++ b/src/auth/types.ts
@@ -38,6 +38,19 @@ export interface SessionAuthConfig {
      * which re-bootstraps `/session`, then retries the original request once.
      */
     refreshOn?: number[];
+    /**
+     * When true, drops the base strategy's headers (e.g. `Authorization: Basic …`) from the
+     * returned `AuthSession` after the `/session` handshake completes. The base headers are
+     * still sent to the session endpoint itself; only post-handshake API calls carry just
+     * cookies + `headerMirror` headers. Required for stateful backends (e.g. Spring Security
+     * with 2FA) where re-presenting the base credentials on every call re-triggers auth filters
+     * and invalidates the active session.
+     *
+     * Drops *all* headers contributed by the base strategy, not just `Authorization`. If a
+     * custom base strategy contributes non-auth headers you need to keep, write a custom
+     * `AuthStrategy` instead of using this flag.
+     */
+    dropBaseHeaders?: boolean;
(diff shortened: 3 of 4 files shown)


garretpremo (commented):

The `dropBaseHeaders` flag is a clean, surgical opt-in: handshake unchanged, post-handshake `AuthSession.headers` filtered, both `authenticate()` and `restore()` paths covered, default behavior preserved. All four acceptance criteria from #106 are met, the new tests pin the tricky bits (onChallenge retry, `restore()` cache rehydration, end-to-end `resolveRequestHeaders`), and CI is green.

<details>

- The TSDoc on `dropBaseHeaders` (`src/auth/types.ts`) explicitly calls out that *all* base headers are dropped — not just `Authorization`. Good caveat for future readers who layer Bearer + session.
- `restore()` mirrors the same filter as `authenticate()`. Without this, a process restart with a cached session on disk would re-introduce the dropped header and silently undo the fix — worth a mention in the PR body, and the test `restore() omits base headers from returned session when enabled` pins it.

</details>

garretpremo (commented):

Approved.

- The TSDoc on `dropBaseHeaders` (`src/auth/types.ts`) explicitly calls out that *all* base headers are dropped — not just `Authorization`. Good caveat for future readers who layer Bearer + session.
- `restore()` mirrors the same filter as `authenticate()`. Without this, a process restart with a cached session on disk would re-introduce the dropped header and silently undo the fix — worth a mention in the PR body, and the test `restore() omits base headers from returned session when enabled` pins it.

PR #110, opened 2026-05-07
feat: project-local CLI aliases for renaming verbose generated commands (#109)

### LINKED ISSUE #109: CLI: support project-local aliases for renaming verbose generated commands
## Summary

Generated leaf-command names come straight from the OpenAPI `operationId`, which spec authors pick for global uniqueness across the spec, not for CLI ergonomics. A spec with an operationId like `getCustomerOrderSummary` for `GET /customers/{customerId}/orderSummary` lands as `mycli customers get-customer-order-summary <id>` — accurate, but verbose and awkward to type. Inconsistencies in operationId style across a single spec (e.g. `getOrderList` vs `orderList` for closely related endpoints) compound the problem.

Today the only way to get short, intuitive command paths is to write a custom command in `.apijack/commands/` that delegates to the generated client. That works for a few high-traffic endpoints but not at scale, and it conflates two concerns — re-aliasing an existing command and adding new behavior — that are otherwise separate.

## Proposal

A project-local `.apijack/aliases.json` (alongside `config.json`) mapping a typed alias to the canonical command path. Flat-map shape, alias on the left, expansion on the right:

```json
{
  "customers list":     "customers get-all-customers",
  "customers summary":  "customers get-customer-order-summary",
  "cs":                 "customers get-customer-order-summary"
}
```

### Resolution semantics

- Implemented as argv rewriting at the CLI bootstrap, before Commander parses. Longest-prefix match on the leading argv tokens. `mycli customers summary 42 --foo bar` → `mycli customers get-customer-order-summary 42 --foo bar`.
- Single-token aliases (`cs`) work the same as multi-token (`customers summary`) — one uniform mechanism.
- Trailing args (positional and flags) are appended verbatim.
- An alias that shadows a real command path emits a startup warning and the real command keeps winning.
- Optional `~/.apijack/aliases.json` as a global fallback for personal shortcuts; project-local takes precedence on conflict.

## Out of scope

- **Routine engine.** Routines should reference canonical command names so a team member retuning their aliases doesn't silently break shared YAML. CLI-only.
- **MCP tool names.** MCP exposes canonical names to the model for the same reason.
- **`--help` output.** Leave generated help untouched in v1; printing aliases under a separate "Aliases" section is a nice follow-up but not required.

## Acceptance criteria

- [ ] `.apijack/aliases.json` is discovered alongside `config.json` and parsed at CLI bootstrap.
- [ ] Argv with a leading alias is rewritten to the canonical path before Commander parses; remaining args are passed through unchanged.
- [ ] Longest-prefix match wins (so `customers summary` beats `customers` when both are aliased).
- [ ] An alias that collides with a real command path emits a startup warning and the real command is invoked.
- [ ] An invalid `aliases.json` (malformed JSON, expansion that resolves to a non-existent command) emits a clear error and the CLI continues without aliases applied.
- [ ] Routine engine and MCP tool resolution are unaffected (canonical names only).
- [ ] Brief docs entry under Project Mode (or its own wiki page) covering format, precedence, and conflict rules.

## Notes

- **Alternative considered:** per-leaf renames via Commander's native `.command(...).alias(...)`. Cleaner for single-token in-group renames but doesn't support cross-group shortcuts (`cs` → `customers get-customer-order-summary`). Argv rewriting handles both with one mechanism.
- **Possible v2 extension:** allow alias values to embed flags (`"customers active": "customers list --status active"`), turning aliases into lightweight curried commands. Deferrable; out of scope for the initial cut.

garretpremo:
Closes #109

## Summary

Generated leaf-command names come straight from the OpenAPI `operationId`, which spec authors pick for global uniqueness, not CLI ergonomics. A spec with `getCustomerOrderSummary` lands as `mycli customers get-customer-order-summary <id>` — accurate, but verbose to type and inconsistent across endpoints.

This PR adds `.apijack/aliases.json` — a flat JSON map from typed aliases to canonical command paths — so users can rename generated commands without writing a custom `.apijack/commands/` module. The mechanism is a pre-Commander argv rewrite: `mycli cs 42 --foo bar` becomes `mycli customers get-customer-order-summary 42 --foo bar`. Single-token (`cs`) and multi-token (`customers summary`) aliases use the same path. Routines and MCP tool resolution stay on canonical names, since shared YAML and tool inventories shouldn't fork per developer.

### New `src/aliases.ts`

- **`loadAliases(configDir, cliName)`** — reads `<configDir>/aliases.json` (project-local) and `~/.<cliName>/aliases.json` (global), merges with project-local taking precedence. Reports malformed JSON, non-object roots, non-string values, and empty keys/values as discrete errors; CLI continues without the offending file's entries.
- **`collectCommandPaths(program)`** — walks the Commander tree post-build to produce a `Set<string>` of canonical paths used for shadow detection and expansion validation.
- **`rewriteArgv(args, aliases, realPaths)`** — full validated rewrite. Aliases that shadow a real command path are skipped with a warning; aliases with expansions that don't resolve to a known command are skipped with an error. Longest-prefix wins (token-count first, char-length tiebreak). Trailing args pass through unchanged.
- **`resolveLeadingTokens(args, aliases)`** — best-effort longest-prefix resolution without validation. Used for the three pre-Commander-build reads of `process.argv[2]`/`[3]` (auth-skip detection, plugins-check guard, dry-run heuristic) so aliases like `gen → generate` interact correctly with auth bootstrap before the program tree exists.

### `src/cli-builder.ts` wire-up

```yaml
# Top of run() — load once, pre-resolve leading tokens
const { map: aliasMap } = loadAliases(configDir, cliName);
const effectiveArgv2 = resolveLeadingTokens(process.argv.slice(2), aliasMap)[0];

# Pre-build argv reads now use effectiveArgv2:
isPluginsCheck       # was: process.argv[2] === 'plugins'
cmd / skipAuthCheck  # was: process.argv[2]
isDryRun             # was: process.argv[2] !== 'routine'

# Just before parseAsync — full validated rewrite mutates process.argv
rewriteArgv(process.argv.slice(2), aliasMap, collectCommandPaths(program))
```

### Example `.apijack/aliases.json`

```json
{
  "customers list":     "customers get-all-customers",
  "customers summary":  "customers get-customer-order-summary",
  "cs":                 "customers get-customer-order-summary"
}
```

### CLAUDE.md

New "Command aliases" subsection under **Project Extensions** documenting format, precedence, conflict rules, and the canonical-name-only guarantee for routines and MCP.

## Acceptance criteria

- [x] `.apijack/aliases.json` is discovered alongside `config.json` and parsed at CLI bootstrap.
- [x] Argv with a leading alias is rewritten to the canonical path before Commander parses; remaining args are passed through unchanged.
- [x] Longest-prefix match wins (so `customers summary` beats `customers` when both are aliased).
- [x] An alias that collides with a real command path emits a startup warning and the real command is invoked.
- [x] An invalid `aliases.json` (malformed JSON, expansion that resolves to a non-existent command) emits a clear error and the CLI continues without aliases applied.
- [x] Routine engine and MCP tool resolution are unaffected (canonical names only).
- [x] Brief docs entry under Project Mode (or its own wiki page) covering format, precedence, and conflict rules.

## Test plan

- [x] `bun test` — 918 pass / 0 fail (26 new tests in `tests/aliases.test.ts` covering loader, tree walk, validated rewrite, and best-effort early resolution)
- [x] `bun run lint` — 0 errors
- [x] Manual inspection: aliases `gen → generate`, `r → routine`, `pcheck → plugins check` would no longer trip the auth-prompt or plugins-check guards before the rewrite, thanks to `resolveLeadingTokens`.

## Out of scope (per #109)

- Routine engine alias support — routines reference canonical names so shared YAML doesn't break per-dev.
- MCP tool name aliasing — MCP exposes canonical names to the model.
- `--help` integration — generated help is untouched in this cut.
- Aliases that embed flags (e.g. `"customers active": "customers list --status active"`) — deferrable v2 polish.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -243,6 +243,30 @@ The `.apijack/` directory at a project root is auto-loaded when the CLI runs ins
 | `.apijack/plugins.ts` | Project-level plugin registrations (`default: ApijackPlugin[]` — each entry passed to `cli.use(...)`) |
 | `.apijack/routines/*.yaml` | Routines available via `routine run <name>` |
 | `.apijack/settings.json` | Framework defaults (see below) |
+| `.apijack/aliases.json` | Project-local command aliases (see below) |
+
+### Command aliases
+
+Generated leaf-command names come from the OpenAPI `operationId` and can be verbose (e.g. `customers get-customer-order-summary`). `.apijack/aliases.json` is a flat map from a typed alias to the canonical command path:
+
+```json
+{
+  "customers list":     "customers get-all-customers",
+  "customers summary":  "customers get-customer-order-summary",
+  "cs":                 "customers get-customer-order-summary"
+}
+```
+
+Resolution rules:
+
+- Argv is rewritten at CLI bootstrap, before Commander parses. `mycli cs 42 --foo bar` becomes `mycli customers get-customer-order-summary 42 --foo bar`.
+- Single-token aliases (`cs`) and multi-token aliases (`customers summary`) use the same mechanism.
+- Trailing positional args and flags pass through unchanged.
+- Longest-prefix wins when multiple aliases could match.
+- An alias that shadows a real command path emits a startup warning and the real command keeps winning.
+- An expansion that doesn't resolve to a known command emits a startup error; the CLI continues without that alias applied.
+- A global `~/.<cliName>/aliases.json` is also consulted; project-local entries override global entries on conflict.
+- **Routines and MCP tool resolution use canonical names only.** Aliases are a CLI ergonomics layer; routine YAML and MCP tool names are not affected.
 
 ### Opt-in auth for custom commands and dispatchers
 
--- a/src/aliases.ts
+++ b/src/aliases.ts
@@ -1,0 +1,231 @@
+import { existsSync, readFileSync } from 'fs';
+import { join } from 'path';
+import { homedir } from 'os';
+import type { Command } from 'commander';
+
+export type AliasMap = Record<string, string>;
+
+export interface LoadedAliases {
+    map: AliasMap;
+    errors: string[];
+}
+
+/**
+ * Load alias definitions from `<configDir>/aliases.json` and `~/.<cliName>/aliases.json`.
+ * Project-local entries override global entries on conflict.
+ */
+export function loadAliases(configDir: string, cliName: string): LoadedAliases {
+    const errors: string[] = [];
+    const globalPath = join(homedir(), '.' + cliName, 'aliases.json');
+    const projectPath = join(configDir, 'aliases.json');
+
+    const global = readAliasFile(globalPath, errors);
+    const project = projectPath === globalPath ? {} : readAliasFile(projectPath, errors);
+
+    return { map: { ...global, ...project }, errors };
+}
+
+function readAliasFile(path: string, errors: string[]): AliasMap {
+    if (!existsSync(path)) return {};
+
+    let raw: string;
+    try {
+        raw = readFileSync(path, 'utf-8');
+    } catch (e) {
+        errors.push(`${path}: failed to read (${(e as Error).message})`);
+
+        return {};
+    }
+
+    let parsed: unknown;
+    try {
+        parsed = JSON.parse(raw);
+    } catch (e) {
+        errors.push(`${path}: malformed JSON (${(e as Error).message})`);
+
+        return {};
+    }
+
+    if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) {
+        errors.push(`${path}: expected a JSON object mapping alias to canonical command`);
+
+        return {};
+    }
+
+    const out: AliasMap = {};
+
+    for (const [k, v] of Object.entries(parsed as Record<string, unknown>)) {
+        if (typeof v !== 'string') {
+            errors.push(`${path}: alias "${k}" must map to a string`);
+            continue;
+        }
+
+        const alias = k.trim();
+        const expansion = v.trim();
+
+        if (!alias || !expansion) {
+            errors.push(`${path}: alias keys and values must be non-empty`);
+            continue;
+        }
+
+        out[alias] = expansion;
+    }
+
+    return out;
+}
+
+/**
+ * Best-effort longest-prefix resolution of leading argv tokens through the
+ * alias map, without validating against the real command tree.
+ *
+ * Used for early bootstrap reads of argv[2]/argv[3] (auth-skip detection,
+ * plugins-check guard, dry-run heuristic) that happen before the Commander
+ * tree is built. The full validated rewrite happens later via `rewriteArgv`.
+ *
+ * If an alias key collides with a real command, this helper still applies the
+ * alias — but the late `rewriteArgv` pass will detect the shadow and skip the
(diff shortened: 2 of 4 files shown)


garretpremo (commented):

PR cleanly implements every acceptance criterion from #109. The two-pass argv handling — best-effort `resolveLeadingTokens` for pre-Commander reads (auth-skip, plugins-check, dry-run) and validated `rewriteArgv` immediately before `parseAsync` — is the right shape; comments call out the trade-off honestly. Diagnostics are surgical: malformed JSON, non-object roots, non-string values, empty keys, shadow conflicts, and unresolvable expansions each get distinct messages. Tests cover loader, tree walk, both rewrite passes, longest-prefix tie-breaks, and partial multi-token non-matches. CI green across all platforms.

<details>

- `rewriteArgv` validates the entire alias map on every invocation, so a single typo'd entry will print its error on every CLI call (including `--help`). Acceptable for now since the surface is small, but if alias maps grow, consider gating diagnostics behind a verbose flag or only emitting them when the matching alias is the leading token.
- `loadAliases` deduplicates global vs. project by string-equal path comparison (`projectPath === globalPath`). If `configDir` is reached via a symlink or different normalization than `homedir() + '.' + cliName`, the file would be read twice and the project copy would still override. Edge case; unlikely in practice.
- `process.argv = [...process.argv.slice(0, 2), ...rewrittenArgs]` mutates global state. Standard practice for argv-rewriting CLIs, but worth noting that any code reading `process.argv` after step 13 sees the rewritten form (intentional — Commander needs it — just flagging for future maintainers).
- The shadow-detection check in `rewriteArgv` is correct, but the warning emits every run. If a user intentionally aliases-then-shadows (e.g. while migrating a name), the noise might prompt them to remove the entry rather than fix the shadow. Minor.

</details>

<details>
<summary>Nitpicks</summary>

- `resolveLeadingTokens` and `rewriteArgv` duplicate the longest-prefix sort + token-by-token match loop. A small private helper (`findLongestPrefixMatch(args, aliasKeys)` returning the matched key + tail) would consolidate both, at the cost of one extra function call.
- The PR body uses `\` at one point in `Sub-routines without their own \`plugins:\` block` style — but that's in the existing CLAUDE.md, not introduced here.

</details>

garretpremo (commented):

Approved.

- `rewriteArgv` validates the entire alias map on every invocation, so a single typo'd entry will print its error on every CLI call (including `--help`). Acceptable for now since the surface is small, but if alias maps grow, consider gating diagnostics behind a verbose flag or only emitting them when the matching alias is the leading token.
- `loadAliases` deduplicates global vs. project by string-equal path comparison (`projectPath === globalPath`). If `configDir` is reached via a symlink or different normalization than `homedir() + '.' + cliName`, the file would be read twice and the project copy would still override. Edge case; unlikely in practice.
- `process.argv = [...process.argv.slice(0, 2), ...rewrittenArgs]` mutates global state. Standard practice for argv-rewriting CLIs, but worth noting that any code reading `process.argv` after step 13 sees the rewritten form (intentional — Commander needs it — just flagging for future maintainers).
- The shadow-detection check in `rewriteArgv` is correct, but the warning emits every run. If a user intentionally aliases-then-shadows (e.g. while migrating a name), the noise might prompt them to remove the entry rather than fix the shadow. Minor.

PR #117, opened 2026-07-10
fix: escape strings and sanitize identifiers in codegen (#115)

### LINKED ISSUE #115: Codegen emits invalid TypeScript for real-world specs: unescaped description strings + dotted operationIds (Jira v3)
## Summary

Codegen produces TypeScript that fails to compile when run against the Jira Cloud platform REST API v3 OpenAPI spec. Two independent defects, both in the codegen string/identifier emission:

1. **Unescaped string literals** — OpenAPI `description` text is emitted directly into double-quoted JS string literals without escaping. Any description containing a newline (or a `"`) produces an unterminated/broken string literal.
2. **Invalid identifiers from dotted operationIds** — `operationId` values are emitted verbatim as client method names. Dotted operationIds (e.g. `AddonPropertiesResource.getAddonProperties_get`) become syntactically invalid method definitions and call sites.

Because `cli-builder` imports the generated `commands`/`client` modules inside a `try/catch` that swallows failures (`Generated commands not available …`), the result is silent: `generate` reports success, but **no API commands register** and `<cli> --help` shows only built-in commands.

## Environment

- `@apijack/core` **1.14.1** (npm `latest`)
- Bun 1.3.11, Linux
- Spec: `https://dac-static.atlassian.com/cloud/jira/platform/swagger-v3.v3.json` (OpenAPI 3.0.1 — "The Jira Cloud platform REST API", 420 paths / 619 operations)

## Reproduction

```bash
mkdir jira && cd jira
# package.json with bin -> bin/<cli>.ts importing @apijack/core/bin/apijack.ts
# .apijack.json:
#   { "name": "jirax", "specUrl": "/cloud/jira/platform/swagger-v3.v3.json", "generatedDir": ".apijack/generated" }
# active env url = https://dac-static.atlassian.com
bunx apijack generate            # reports success
bun build .apijack/generated/commands.ts --target=bun   # fails
bun build .apijack/generated/client.ts   --target=bun   # fails
jirax --help                     # no API command groups (silently)
```

## Bug 1 — unescaped description strings

`bun build .apijack/generated/commands.ts`:

```
218 | ption("--sendBulkNotification <value>", "A boolean value that indicates whether to send a bulk change notification when
                                                    ^
error: Unterminated string literal at commands.ts:218:47
```

The Commander `.option(..., "<description>")` is emitted with the raw spec description, which wraps across a newline. In this spec:

- **956** `description` fields contain newlines
- **96** `description` fields contain double-quote characters

So this is pervasive for any verbose spec, not an edge case.

## Bug 2 — dotted operationIds emitted as identifiers

`bun build .apijack/generated/client.ts`:

```
10488 |   async AddonPropertiesResource.getAddonProperties_get(addonKey: string): Promise<PropertyKeys> {
                                       ^
error: Expected "(" but found "."
```

And the corresponding call sites in `commands.ts`:

```ts
const result = await client.AddonPropertiesResource.getAddonProperties_get(addonKey);
const result = await client.DynamicModulesResource.registerModules_post(body);
```

13 operations in this spec have dotted operationIds (Atlassian Connect / app-management resources: `AddonPropertiesResource.*`, `DynamicModulesResource.*`, `AppIssueFieldValueUpdateResource.*`, `MigrationResource.*`, …).

## Suggested fixes

1. **Escape all emitted string literals.** Use `JSON.stringify(description)` (or escape `\`, `"`, `\n`, `\r`, `\t`) wherever spec-sourced text is written into generated source — at minimum option descriptions and any doc comments/strings.
2. **Sanitize operationIds into valid identifiers** for method names (e.g. replace `[^A-Za-z0-9_]` with `_`, ensure it doesn't start with a digit), while keeping the original operationId only where the HTTP path/operation mapping needs it. Apply the same identifier in the client definition and its call sites so they stay in sync.
3. **Optional:** surface a warning instead of silently swallowing a generated-module import error in `cli-builder`, so a broken `generate` is visible rather than presenting as "no commands."

Happy to provide the full generated output or a minimal failing spec excerpt if useful.

garretpremo:
Closes #115

## Summary

apijack's OpenAPI codegen emitted TypeScript that fails to compile for verbose real-world specs (reproduced against the Jira Cloud platform REST API v3, 619 operations). Because `cli-builder` imported the generated modules inside a bare `try/catch`, the failure was silent: `generate` reported success but no API commands registered.

This fixes four independent emission defects — all cases where spec-sourced text or an `operationId`/tag was written into generated source without being escaped or sanitized into a valid JS identifier — and makes the previously-silent import failure visible.

### New shared helpers (`src/codegen/util.ts`)

- `sanitizeIdentifier(name)` — `[^A-Za-z0-9_$]` → `_`, and prefixes `_` when the result starts with a digit or is a reserved word. Deterministic, so every emission site derives the same identifier.
- `emitKey(name)` — returns the name as-is when it is a valid JS identifier, else `JSON.stringify(name)`.
- `sanitizeVar` — hardened to guard reserved words and a leading digit (shared `RESERVED_WORDS` set).

### Bug 1 — unescaped description string literals

The emitters escaped `"` only, never `\n` / `\r` / `\` / `\t`, so any description with a newline split the JS string literal. Replaced the `"${x.replace(/"/g,'\\"')}"` patterns with `${JSON.stringify(x)}` at the three `commands.ts` description sites and the `command-map.ts` description. Also escaped `*/` in the `client.ts` JSDoc builder so a description containing `*/` no longer closes the comment early.

### Bug 2 — dotted operationIds emitted as identifiers

`operationId` values like `AddonPropertiesResource.getAddonProperties_get` were emitted verbatim as method names, call sites, and variable names. Ran them through `sanitizeIdentifier` at every site that must stay in sync:

- `client.ts` — `async <name>(...)` method definition
- `commands.ts` — `client.<name>(...)` call site, the `_<name>` variable and its `configureHelp` use
- `command-map.ts` — the emitted `operationId: "<name>"` **string**, consumed at runtime by `src/routine/dispatcher.ts` as `client[methodName]`. Sanitizing identically here keeps routine dispatch working.

### Bug 3 — reserved-word group variable for untagged operations

An untagged operation is grouped under tag `default` and was emitted as `const default = ...` (a reserved word). `sanitizeVar` now guards reserved words and leading digits, so it emits `const _default = ...`.

### Bug 4 — non-identifier schema property keys emitted unquoted

Keys like Jira's `16x16`/`24x24` were interpolated raw (`16x16?: string;` — a syntax error). Property keys are now emitted via `emitKey(...)` at `types.ts` (interface members and `allOf` inline members) and `util.ts` (`resolveType` inline object literals), producing `"16x16"?: string;`.

### Scoped extra — surface silent import failures (`cli-builder.ts`)

The two bare `catch {}` swallows around generated-module imports now distinguish a genuinely-absent module (`err.code === 'ERR_MODULE_NOT_FOUND'` — expected before the first `generate`, stays silent) from a module that exists but fails to import (a codegen syntax error — now warns to stderr). Routine-only consumers who never generate see no new noise.

## Acceptance criteria

- [x] Generated `commands.ts`, `client.ts`, `types.ts`, and `command-map.ts` are valid TypeScript for descriptions containing newlines/quotes/backslashes/`*/`, dotted operationIds, untagged operations, and non-identifier property keys (e.g. `16x16`).
- [x] Routine dispatch still resolves a sanitized-name client method — the command-map `operationId` string matches the emitted client method name exactly.
- [x] No behavior change for specs that already generate cleanly (existing codegen tests unchanged and passing; committed `.apijack/generated` untouched).
- [x] New codegen unit tests cover each of the four cases.

## Known limitations / out of scope

- **`sanitizeIdentifier` is not injective** — two operationIds differing only by a non-identifier char vs `_` (e.g. `Foo.bar` and `Foo_bar`) collapse to the same method name. No such collision exists in the Jira v3 spec (the dotted ops have distinct prefixes), and the sanitizer stays a deterministic per-site pure function so all three emission sites agree. A de-dup guard is tracked as a follow-up in #118.
- **Dotted operationIds can leak a `.` into a CLI command *name*** on verb collision (the kebab-cased fallback command token at `commands.ts`/`command-map.ts` keeps the `.`). This is pre-existing, cosmetic (the token is a quoted string, not an identifier — no compile error), and out of scope for this fix.

## Test plan

- [x] `bun test` — 928 pass / 0 fail (10 new regression tests in `tests/codegen/invalid-typescript.test.ts`).
- [x] `bun run lint` — 0 errors (pre-existing warnings only).
- [x] New tests feed the generated source through `Bun.Transpiler`, which throws on any syntax error — so a regression fails the test rather than merely differing in text. One test asserts the command-map `operationId` string equals the emitted client method name (the cross-site sync guarantee).

### DIFF
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -213,8 +213,14 @@ export function createCli(options: CliOptions): Cli {
                 const clientMod = await import(resolve(generatedDir, 'client'));
                 ApiClientClass = clientMod.ApiClient;
             }
-        } catch {
-            // Generated commands not available — routines that don't dispatch to API methods still work.
+        } catch (err) {
+            // A genuinely-absent module is expected before the first `generate`
+            // (routines that don't dispatch to API methods still work) — stay
+            // silent. A module that exists but fails to import (e.g. a codegen
+            // syntax error) is a real problem — surface it to stderr.
+            if ((err as { code?: string }).code !== 'ERR_MODULE_NOT_FOUND') {
+                console.warn(`[apijack] Generated module failed to import: ${(err as Error).message}`);
+            }
         }
 
         // 5. Build CliContext with lazy session.
@@ -573,8 +579,13 @@ export function createCli(options: CliOptions): Cli {
                     const clientMod = await import(resolve(generatedDir, 'client'));
                     ApiClientClass = clientMod.ApiClient;
                 }
-            } catch {
-                // Generated commands not available — consumer hasn't run `generate` yet
+            } catch (err) {
+                // A genuinely-absent module is expected before the consumer runs
+                // `generate` — stay silent. A module that exists but fails to
+                // import (e.g. a codegen syntax error) is a real problem — surface it.
+                if ((err as { code?: string }).code !== 'ERR_MODULE_NOT_FOUND') {
+                    console.warn(`[apijack] Generated module failed to import: ${(err as Error).message}`);
+                }
             }
 
             // 7. Detect request-preview output modes
--- a/src/codegen/client.ts
+++ b/src/codegen/client.ts
@@ -1,6 +1,6 @@
 import type { OpenApiOperation, OpenApiSchema } from './openapi-types';
 import { HTTP_METHODS } from './openapi-types';
-import { schemaToTsType, refToName, resolveType, resolveResponseType } from './util';
+import { schemaToTsType, refToName, resolveType, resolveResponseType, sanitizeIdentifier } from './util';
 
 /**
  * Generate an ApiClient class from OpenAPI paths.
@@ -147,22 +147,24 @@ export function generateClient(
                 }
             }
 
-            // Emit JSDoc
-            if (docLines.length === 1) {
-                methodLines.push(`  /** ${docLines[0]} */`);
-            } else {
-                methodLines.push(`  /** ${docLines[0]}`);
+            // Emit JSDoc — escape `*/` so a description doesn't close the comment early
+            const safeDocLines = docLines.map(l => l.replace(/\*\//g, '*\\/'));
 
-                for (let i = 1; i < docLines.length; i++) {
-                    if (i === docLines.length - 1) {
-                        methodLines.push(`   * ${docLines[i]} */`);
+            if (safeDocLines.length === 1) {
+                methodLines.push(`  /** ${safeDocLines[0]} */`);
+            } else {
+                methodLines.push(`  /** ${safeDocLines[0]}`);
+
+                for (let i = 1; i < safeDocLines.length; i++) {
+                    if (i === safeDocLines.length - 1) {
+                        methodLines.push(`   * ${safeDocLines[i]} */`);
                     } else {
-                        methodLines.push(`   * ${docLines[i]}`);
+                        methodLines.push(`   * ${safeDocLines[i]}`);
                     }
                 }
             }
 
-            methodLines.push(`  async ${op.operationId}(${args.join(', ')}): Promise<${returnType}> {`);
+            methodLines.push(`  async ${sanitizeIdentifier(op.operationId)}(${args.join(', ')}): Promise<${returnType}> {`);
             methodLines.push(`    return this.request("${method.toUpperCase()}", ${pathTemplate}${optsArg}) as Promise<${returnType}>;`);
             methodLines.push('  }');
             methodLines.push('');
--- a/src/codegen/command-map.ts
+++ b/src/codegen/command-map.ts
@@ -1,6 +1,6 @@
 import type { OpenApiOperation, OpenApiSchema } from './openapi-types';
 import { HTTP_METHODS } from './openapi-types';
-import { normalizeTag, resolveSchemaProps } from './util';
+import { normalizeTag, resolveSchemaProps, sanitizeIdentifier } from './util';
 
 /**
  * Generate a command map that maps CLI command paths to their
@@ -138,12 +138,15 @@ export function generateCommandMap(
                         ? `${groupName} ${cmdName}`
                         : `${groupName} ${resourceName} ${cmdName}`;
 
-                const descPart = cmd.summary ? `, description: "${cmd.summary.replace(/"/g, '\\"')}"` : '';
+                const descPart = cmd.summary ? `, description: ${JSON.stringify(cmd.summary)}` : '';
                 const bodyPart = cmd.bodyFields.length > 0
                     ? `, bodyFields: ${JSON.stringify(cmd.bodyFields)}`
                     : '';
+                // Sanitize identically to the client method name so the
+                // dispatcher's client[operationId] lookup resolves at runtime.
+                const methodName = sanitizeIdentifier(cmd.operationId);
                 entries.push(
-                    `  "${cmdPath}": { operationId: "${cmd.operationId}", pathParams: [${cmd.pathParams.map(p => `"${p}"`).join(', ')}], queryParams: [${cmd.queryParams.map(p => `"${p}"`).join(', ')}], hasBody: ${cmd.hasBody}${bodyPart}${descPart} },`,
+                    `  "${cmdPath}": { operationId: "${methodName}", pathParams: [${cmd.pathParams.map(p => `"${p}"`).join(', ')}], queryParams: [${cmd.queryParams.map(p => `"${p}"`).join(', ')}], hasBody: ${cmd.hasBody}${bodyPart}${descPart} },`,
                 );
             }
         }
--- a/src/codegen/commands.ts
+++ b/src/codegen/commands.ts
@@ -9,6 +9,7 @@ import {
     normalizeTag,
     resolveSchemaProps,
     sanitizeVar,
+    sanitizeIdentifier,
     capitalize,
 } from './util';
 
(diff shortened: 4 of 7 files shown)


garretpremo (commented):

All four acceptance criteria are met and no blocking issues were found. The `JSON.stringify` escaping, shared `sanitizeIdentifier`/`emitKey` helpers, and the cross-site sync between the client method name and the command-map `operationId` string are all correct — the sync test in `tests/codegen/invalid-typescript.test.ts` pins the actual runtime contract used by `src/routine/dispatcher.ts`. The `cli-builder` change was verified empirically: in Bun, a missing module throws with `code: "ERR_MODULE_NOT_FOUND"` while a module with a syntax error throws a `BuildMessage` with `code: undefined`, so broken codegen now warns and "not generated yet" stays silent, exactly as intended. Full suite passes locally (928/928) and CI is green on both platforms.

<details>

- **Same-class emission bugs remain at parameter/property-name sites (pre-existing, follow-up material).** Spec-sourced names are still interpolated unsanitized where they must be identifiers or bare object keys, and each was empirically confirmed to emit invalid TypeScript today: path-param names as method args (`src/codegen/client.ts:59`) and as local consts/call args (`src/codegen/commands.ts:318,328`); query-param names in the `params` object literal and `opts.<name>` access (`src/codegen/commands.ts:380`) and in the client `params?: { ... }` type (`src/codegen/client.ts:73`); body-prop keys like `16x16` in `opts.${bp.camelName}` (`src/codegen/commands.ts:362`) — the exact Jira key shape bug 4 fixed for `types.ts`. E.g. a hyphenated param name (`page-size`) or bracket-style JSON:API param (`filter[status]`) breaks generation. The query-param fix needs Commander's flag-camelCasing semantics thought through, so it deserves its own issue rather than growing this PR.
- **`buildJsDoc` escapes `*/` in the description only; tag values leak** (`src/codegen/util.ts:243-301`). Confirmed: a schema `pattern: "^.*/end$"` makes `generateTypes` emit `/** @pattern ^.*/end$ */`, which closes the comment early — a syntax error. Pre-existing (this PR fixed the same class in the client JSDoc builder); the fix is applying the same `*/` escape to the tag lines. Good candidate for the same follow-up issue.
- **`sanitizeIdentifier`'s reserved-word guard renames a previously-valid operationId.** Reserved words are legal as class method names and property accesses, so an operationId of exactly `delete` used to emit working `async delete()` and now emits `_delete`. Verified the rename is self-consistent (command-map says `_delete` too, so CLI and routine dispatch work) and the behavior is documented in the PR body and JSDoc — accepted posture, noting only that a consumer calling the generated `ApiClient` directly would see the rename after regenerating.

</details>

<details>
<summary>Nitpicks</summary>

- `sanitizeIdentifier` can collide distinct operationIds (`a.b` and `a_b` both → `a_b`): the client class gets duplicate methods (last wins, silently) — a runtime wrong-answer rather than a compile error. `sanitizeTypeName` already uses a `.` → `__` mapping for the analogous problem; worth adopting or documenting the collision posture in the JSDoc if a spec ever hits it.
- The "generateCommands emits a valid call site" test would not have failed pre-fix: `client.AddonPropertiesResource.getAddonProperties_get(...)` parses as a property chain, and the fixture op has no variant props so no `_<id>` const is emitted. The real regression pin is the method-name match test below it; adding variant props to the dotted-operationId fixture would make the parse test bite too.
- When verbs collide, the kebab-cased raw operationId can retain a dot in the CLI command name (`addon-properties-resource.get-addon-properties_get`). It is inside a string literal, both emitters derive it identically, and Commander accepts it — just slightly odd UX for dotted-operationId specs.

</details>

garretpremo (commented):

Approved.

- **Same-class emission bugs remain at parameter/property-name sites (pre-existing, follow-up material).** Spec-sourced names are still interpolated unsanitized where they must be identifiers or bare object keys, and each was empirically confirmed to emit invalid TypeScript today: path-param names as method args (`src/codegen/client.ts:59`) and as local consts/call args (`src/codegen/commands.ts:318,328`); query-param names in the `params` object literal and `opts.<name>` access (`src/codegen/commands.ts:380`) and in the client `params?: { ... }` type (`src/codegen/client.ts:73`); body-prop keys like `16x16` in `opts.${bp.camelName}` (`src/codegen/commands.ts:362`) — the exact Jira key shape bug 4 fixed for `types.ts`. E.g. a hyphenated param name (`page-size`) or bracket-style JSON:API param (`filter[status]`) breaks generation. The query-param fix needs Commander's flag-camelCasing semantics thought through, so it deserves its own issue rather than growing this PR.
- **`buildJsDoc` escapes `*/` in the description only; tag values leak** (`src/codegen/util.ts:243-301`). Confirmed: a schema `pattern: "^.*/end$"` makes `generateTypes` emit `/** @pattern ^.*/end$ */`, which closes the comment early — a syntax error. Pre-existing (this PR fixed the same class in the client JSDoc builder); the fix is applying the same `*/` escape to the tag lines. Good candidate for the same follow-up issue.
- **`sanitizeIdentifier`'s reserved-word guard renames a previously-valid operationId.** Reserved words are legal as class method names and property accesses, so an operationId of exactly `delete` used to emit working `async delete()` and now emits `_delete`. Verified the rename is self-consistent (command-map says `_delete` too, so CLI and routine dispatch work) and the behavior is documented in the PR body and JSDoc — accepted posture, noting only that a consumer calling the generated `ApiClient` directly would see the rename after regenerating.

PR #119, opened 2026-07-10
fix: coerce object-body props by resolved schema type (#116)

garretpremo:
Closes #116

## Summary

The generated CLI's object-body branch shipped every body property verbatim (`obj["x"] = opts.x`). Commander option values are always strings, so array/object props reached strict APIs as raw JSON strings (`{"line_items": "[{…}]"}` → 400 `"expected": "array", "received": "string"`), and number/boolean props went out as `"42"`/`"true"`. The primitive-body branch just above already coerced; the per-prop object branch never got the same treatment.

This adds a `coerce` kind to each `BodyProp`, computed at construction from the resolved schema, and emits per-kind coercion in the generated action.

### Coercion classification (`src/codegen/util.ts`)

`schemaToTsType` collapses arrays and inline objects to `unknown`, so keying coercion on `BodyProp.type` (as the issue originally proposed) can't work. Instead a new `bodyPropCoercion(prop, schemas, enumValues)` helper classifies at `BodyProp` construction time in `resolveSchemaProps`, where the raw and `$ref`-resolvable schema is available:

- `integer` / `number` → `number`
- `boolean` → `boolean`
- `string`, and scalar enums (`enumValues` present) → `string`
- `array` / `object` / `$ref`-to-object / otherwise-`unknown` → `json`

It resolves a single `$ref` so a ref to a string enum stays `string` (no parse) while a ref to an object becomes `json`.

### Emission (`src/codegen/commands.ts`)

The object-body branch now switches on `bp.coerce`:

```ts
// number
if (opts.count !== undefined) {
  const n = Number(opts.count);
  if (Number.isNaN(n)) throw new Error("--count expects a number");
  obj["count"] = n;
}
// boolean
if (opts.flagged !== undefined) obj["flagged"] = opts.flagged === "true";
// json
if (opts.lineItems !== undefined) {
  try { obj["lineItems"] = JSON.parse(opts.lineItems as string); }
  catch { throw new Error("--lineItems expects JSON, e.g. '[{\"id\":...}]'"); }
}
// string — unchanged passthrough
if (opts.name !== undefined) obj["name"] = opts.name;
```

### New field (`src/codegen/openapi-types.ts`)

`BodyProp.coerce: 'number' | 'boolean' | 'json' | 'string'` (required; the sole construction site is `resolveSchemaProps`).

## Scope decisions (from triage, resolved autonomously)

- **Query params — out of scope.** The client serializes params via `String(v)` into a query string, which carries strings by definition; there is no coercion bug there. Array-typed query params (OpenAPI `style`/`explode`) are a separate concern not addressed here.
- **Comma-split fallback for `string[]` — dropped.** Array element type is lost at `BodyProp` construction (everything array-shaped becomes `json`), so a CSV fallback couldn't be distinguished from a JSON array and would corrupt the primary case. `JSON.parse` only.

### UX note — two array input styles, by shape

There is an intentional divergence in how array input is accepted, driven by where the array sits:

- **Array-typed *object properties*** (this PR) now require JSON: `--tags '["a","b"]'`.
- **Top-level primitive array *bodies*** (the pre-existing primitive-body branch, unchanged) still comma-split: `-B a,b`.

The property branch can't reliably comma-split because element type is erased at classification time (see above), so JSON is the single unambiguous form. Numeric/boolean *enums* (including `$ref`-to-enum) are classified by member type, so a numeric enum coerces with `Number()` rather than shipping `"2"`.

## Acceptance criteria

- [x] Object-body `number`/`integer` props coerced via `Number(...)`, erroring on `NaN`.
- [x] Object-body `boolean` props coerced via `=== "true"`.
- [x] Object-body array/object/complex props parsed via `JSON.parse` with a clear per-flag error naming the flag and showing a JSON example.
- [x] `string` (and scalar-enum) props unchanged.
- [x] Unit coverage on `generateCommands` output for each coercion kind — first direct coverage of the object-body coercion.
- [x] No change to query-param handling.

## Test plan

- [x] `bun test` — 947 pass / 0 fail (19 new in `tests/codegen/body-coercion.test.ts`)
- [x] `bun run lint` — 0 errors
- [x] New tests both transpile the emitted source (catches escaping bugs in the JSON error message) and execute the extracted coercion block: `'42'`→`42`, `'true'`→`true`, `'[{"id":"a","done":true}]'`→real array, non-numeric/invalid-JSON throw the per-flag error, string passthrough unchanged, `$ref`→string-enum stays string, `$ref`→object becomes JSON.

### DIFF
--- a/src/codegen/commands.ts
+++ b/src/codegen/commands.ts
@@ -358,9 +358,37 @@ export function generateCommands(
                         lines.push('      const obj: Record<string, unknown> = {};');
 
                         for (const bp of cmd.bodyProps) {
-                            lines.push(
-                                `      if (opts.${bp.camelName} !== undefined) obj["${bp.name}"] = opts.${bp.camelName};`,
-                            );
+                            // Commander flag values are always strings — coerce
+                            // per the prop's resolved schema (see issue #116).
+                            switch (bp.coerce) {
+                                case 'number':
+                                    lines.push(
+                                        `      if (opts.${bp.camelName} !== undefined) {`,
+                                        `        const n = Number(opts.${bp.camelName});`,
+                                        `        if (Number.isNaN(n)) throw new Error("--${bp.cliFlag} expects a number");`,
+                                        `        obj["${bp.name}"] = n;`,
+                                        '      }',
+                                    );
+                                    break;
+                                case 'boolean':
+                                    lines.push(
+                                        `      if (opts.${bp.camelName} !== undefined) obj["${bp.name}"] = opts.${bp.camelName} === "true";`,
+                                    );
+                                    break;
+                                case 'json':
+                                    lines.push(
+                                        `      if (opts.${bp.camelName} !== undefined) {`,
+                                        `        try { obj["${bp.name}"] = JSON.parse(opts.${bp.camelName} as string); }`,
+                                        `        catch { throw new Error("--${bp.cliFlag} expects JSON, e.g. '[{\\"id\\":...}]'"); }`,
+                                        '      }',
+                                    );
+                                    break;
+                                default:
+                                    // string / scalar enum — passthrough
+                                    lines.push(
+                                        `      if (opts.${bp.camelName} !== undefined) obj["${bp.name}"] = opts.${bp.camelName};`,
+                                    );
+                            }
                         }
 
                         if (cmd.bodyIsArray) {
--- a/src/codegen/openapi-types.ts
+++ b/src/codegen/openapi-types.ts
@@ -92,6 +92,13 @@ export const HTTP_METHODS: HttpMethod[] = [
 export interface BodyProp {
     name: string;
     type: string;
+    /**
+     * How the always-string CLI flag value is converted before it goes into
+     * the request body. Computed from the resolved schema at construction time
+     * (see resolveSchemaProps): 'number' → Number(), 'boolean' → === "true",
+     * 'json' → JSON.parse (arrays/objects/complex), 'string' → passthrough.
+     */
+    coerce: 'number' | 'boolean' | 'json' | 'string';
     cliFlag: string; // kebab-case for CLI, e.g. "first-name"
     camelName: string; // camelCase for Commander opts, e.g. "firstName"
     enumValues?: (string | number | boolean | null)[];
--- a/src/codegen/util.ts
+++ b/src/codegen/util.ts
@@ -218,6 +218,65 @@ export function schemaToTsType(schema: OpenApiSchema): string {
 }
 
 /**
+ * Classify how a body property's always-string CLI flag value must be coerced
+ * before it goes into the request body. `schemaToTsType` collapses arrays and
+ * inline objects to `unknown`, so this resolves the raw (and `$ref`-able)
+ * schema instead: a `$ref` to a string enum stays a string (no parse) while a
+ * `$ref` to an object becomes JSON.
+ */
+export function bodyPropCoercion(
+    prop: OpenApiSchema,
+    schemas: Record<string, OpenApiSchema>,
+    enumValues: (string | number | boolean | null)[] | undefined,
+): 'number' | 'boolean' | 'json' | 'string' {
+    // Scalar enums arrive as plain strings — never JSON. Classify by member
+    // type so a numeric/boolean enum still coerces (a string enum passes
+    // through). null members are ignored (nullable enums).
+    if (enumValues) {
+        const nonNull = enumValues.filter(v => v !== null);
+
+        if (nonNull.length > 0 && nonNull.every(v => typeof v === 'number')) return 'number';
+
+        if (nonNull.length > 0 && nonNull.every(v => typeof v === 'boolean')) return 'boolean';
+
+        return 'string';
+    }
+
+    // Resolve $refs (transitively, cycle-guarded) so classification sees the
+    // underlying type — a $ref-to-object becomes JSON, a $ref-to-string stays
+    // a string.
+    let resolved = prop;
+    const seen = new Set<string>();
+
+    while (resolved.$ref) {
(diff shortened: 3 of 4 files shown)


garretpremo (commented):

The implementation matches the triaged design and the tests are strong (emitted-source parse checks plus runtime execution of the extracted coercion block). One regression blocks the merge: `bodyPropCoercion` doesn't normalize OAS 3.1 nullable type forms, so nullable *string* props flip from working passthrough to a hard `expects JSON` error.

**Blocking:**

- **OAS 3.1 nullable scalar props regress from working passthrough to throwing on bare input.** `bodyPropCoercion` switches on `resolved.type` (`src/codegen/util.ts:265`) without normalizing the OAS 3.1 array form, so `{ type: ["string", "null"] }` falls to `default:` → `json`, and the generated action does `JSON.parse` on the flag value. `--name hello` then throws `--name expects JSON` where the pre-PR passthrough sent it correctly. Verified against this branch: `gen({ name: { type: ['string','null'] } })` emits `JSON.parse(opts.name ...)`, no passthrough. The same happens for the `anyOf: [{type:"string"},{type:"null"}]` nullable pattern (ubiquitous in Pydantic/FastAPI-generated specs). Nullable number/boolean props similarly land on `json` instead of their scalar coercion, though those were already broken pre-PR so only the string case is a strict regression. The codebase already normalizes OAS 3.1 type arrays in `resolveType` (`src/codegen/util.ts:42-44`) and ships a dedicated "OAS 3.1 Features" CI suite — `bodyPropCoercion` should do the same: filter `"null"` from an array `type` and classify by the single remaining member (and ideally classify a single-non-null-variant `anyOf`/`oneOf` the same way). Please add nullable-prop cases to `tests/codegen/body-coercion.test.ts` when fixing. For reference, OAS 3.0 `nullable: true` is unaffected (`type` stays a plain string) — verified passthrough still emits.

<details>

- `Number("")` is `0`, so `--count ""` silently sends `0` instead of tripping the NaN guard. An explicit empty-string check (or `opts.x === '' || Number.isNaN(n)`) would match the "error on invalid number" intent. Same gap exists in the primitive-body branch, so consistency is a fair defense for leaving it.
- Boolean coercion via `=== "true"` means `--flagged yes` or `--flagged TRUE` silently becomes `false`. This mirrors the primitive-body branch precedent, so it's consistent — but a validation error on anything other than `true`/`false` would be kinder, and could be a follow-up for both branches together.
- Free-form props (empty schema `{}`) now classify as `json`, so a bare string value must be quoted (`--meta '"hello"'`). This follows the triage decision ("otherwise-unknown → json") and is the right call for round-tripping objects/numbers, just noting the UX change is user-visible.

</details>

<details>
<summary>Nitpicks</summary>

- The new emission lines interpolate `bp.cliFlag` and `bp.name` raw into generated string literals (e.g. `throw new Error("--${bp.cliFlag} expects a number")`). A prop name containing `"` would break the emitted source. The pre-existing assignment line had the same exposure and #115 added escaping helpers for exactly this class of problem — worth routing these through the same escaping when convenient.

</details>

garretpremo (commented):

Re-review after the previous "changes requested" round. The blocking OAS 3.1 nullable regression is fixed: 7c5bc66 normalizes the array `type` form and single-non-null-variant `anyOf`/`oneOf` to the sole non-null member, matching `resolveType`'s existing 3.1 handling, with nullable-prop cases added to the test suite. Verified against this branch: `{ type: ["string","null"] }`, `anyOf: [{type:"string"},{type:"null"}]`, and `anyOf: [$ref → string, null]` all emit passthrough; `oneOf: [{type:"integer"},{type:"null"}]` emits `Number()` coercion; `{ type: ["array","null"] }` emits `JSON.parse`; OAS 3.0 `nullable: true` unaffected. All six triage acceptance criteria are met, 951 tests pass locally, and all 9 CI checks are green. No blocking issues.

<details>

- The three observations from the first review still stand as potential follow-ups: `Number("")` is `0` so `--count ""` silently sends `0` past the NaN guard; boolean coercion via `=== "true"` silently maps `yes`/`TRUE` to `false`; and free-form `{}` props now require quoted JSON strings (`--meta '"hello"'`). All are consistent with the primitive-body branch precedent and/or the triage design, so none block.
- A union with two or more non-null variants (e.g. `oneOf: [{type:"string"},{type:"integer"}]`) classifies as `json`, so a bare string value must be quoted. Consistent with the agreed "otherwise-unknown → json" rule; just noting it's user-visible for specs that use scalar unions.

</details>

<details>
<summary>Nitpicks</summary>

- The new `anyOf`/`oneOf` branch recurses via `return bodyPropCoercion(nonNull[0], schemas, undefined)` without threading the `seen` cycle guard, so a degenerate alias cycle expressed through `anyOf` (e.g. `A: { anyOf: [{ $ref: "#/components/schemas/A" }, { type: "null" }] }`) would recurse until stack overflow at generation time. The `while` loop already guards plain `$ref` cycles; passing `seen` through the recursion (or capping depth) would close the gap. Requires a pathological spec, so nitpick only.
- The `bp.cliFlag`/`bp.name` raw interpolation nitpick from the first review remains open — same exposure class as #115, worth routing through the escaping helpers when convenient.

</details>

garretpremo (commented):

Approved.

- The three observations from the first review still stand as potential follow-ups: `Number("")` is `0` so `--count ""` silently sends `0` past the NaN guard; boolean coercion via `=== "true"` silently maps `yes`/`TRUE` to `false`; and free-form `{}` props now require quoted JSON strings (`--meta '"hello"'`). All are consistent with the primitive-body branch precedent and/or the triage design, so none block.
- A union with two or more non-null variants (e.g. `oneOf: [{type:"string"},{type:"integer"}]`) classifies as `json`, so a bare string value must be quoted. Consistent with the agreed "otherwise-unknown → json" rule; just noting it's user-visible for specs that use scalar unions.

PR #120, opened 2026-07-11
fix: de-duplicate colliding operationIds in codegen (#118)

garretpremo:
Closes #118

## Summary

`sanitizeIdentifier` (src/codegen/util.ts) maps any `[^A-Za-z0-9_$]` character to `_`, so it is not injective: two distinct operationIds that differ only by a non-identifier character vs. an underscore collapse to the same method name (e.g. both `Foo.bar` and `Foo_bar` → `Foo_bar`). When a single spec contains both, `generateClient` emits two identical `async Foo_bar(` methods (the second shadows the first) and both `generateCommandMap` entries carry `operationId: "Foo_bar"` — so one CLI command dispatches to the wrong operation at runtime (`client[methodName]`).

This replaces the three independent `sanitizeIdentifier(op.operationId)` call sites with a single shared, deterministic name map that de-duplicates on collision, keeping the three emission sites in sync by construction.

### New `buildMethodNameMap(paths)` (src/codegen/util.ts)

Builds a `Map<rawOperationId, uniqueMethodName>` by iterating in the same `Object.entries(paths)` × `HTTP_METHODS` order all three generators already use. For each operation it sanitizes the operationId and, if that name is already taken, appends a numeric suffix (`_2`, `_3`, …) — checking a `used` set each time so a suffixed candidate that happens to equal a real operationId's name (e.g. a genuine `Foo_bar_2` op) is skipped rather than reused. Look-ups are keyed by the raw operationId, so iteration order only affects suffix assignment, not correctness.

### Shared lookup at all three emission sites

`generateClient`, `generateCommands`, and `generateCommandMap` each build the map once and replace their `sanitizeIdentifier(op.operationId)` / `sanitizeIdentifier(cmd.operationId)` call with `methodNames.get(...)!`. Because the map is keyed by raw operationId and built with identical iteration order in each generator, the client method name, the command-map `operationId` string, and the command's local method reference all resolve to the same unique name.

`sanitizeIdentifier` itself is unchanged (still exported, now also consumed internally by `buildMethodNameMap`).

## Acceptance criteria

- [x] A spec containing both `Foo.bar` and `Foo_bar` operationIds generates two distinct client method names.
- [x] The command-map `operationId` string for each still matches its emitted client method name (cross-site sync preserved).
- [x] Regression test covering the collision case.

## Test plan

- [x] `bun test` — 955 pass / 0 fail
- [x] `bun run lint` — 0 errors (pre-existing warnings only)
- [x] New regression tests in `tests/codegen/invalid-typescript.test.ts` (`#118` block):
  - two-way collision (`Foo.bar` vs `Foo_bar`) → two distinct client methods, both command-map operationIds distinct and matching a real method
  - three-way collision (`Foo.bar` / `Foo_bar` / `Foo-bar`) → `_2` / `_3` suffixes
  - suffix-vs-real-name collision (`Foo_bar_2` real op present) → `used`-set bumps the collision pair past the real name; all three distinct and command-map sync holds

### DIFF
--- a/src/codegen/client.ts
+++ b/src/codegen/client.ts
@@ -1,6 +1,6 @@
 import type { OpenApiOperation, OpenApiSchema } from './openapi-types';
 import { HTTP_METHODS } from './openapi-types';
-import { schemaToTsType, refToName, resolveType, resolveResponseType, sanitizeIdentifier } from './util';
+import { schemaToTsType, refToName, resolveType, resolveResponseType, buildMethodNameMap } from './util';
 
 /**
  * Generate an ApiClient class from OpenAPI paths.
@@ -30,6 +30,7 @@ export function generateClient(
     }
 
     const methodLines: string[] = [];
+    const methodNames = buildMethodNameMap(paths);
 
     for (const [path, methods] of Object.entries(paths)) {
         const pathLevelParams: NonNullable<OpenApiOperation['parameters']> = (methods as Record<string, unknown>).parameters as NonNullable<OpenApiOperation['parameters']> || [];
@@ -164,7 +165,7 @@ export function generateClient(
                 }
             }
 
-            methodLines.push(`  async ${sanitizeIdentifier(op.operationId)}(${args.join(', ')}): Promise<${returnType}> {`);
+            methodLines.push(`  async ${methodNames.get(op.operationId)!}(${args.join(', ')}): Promise<${returnType}> {`);
             methodLines.push(`    return this.request("${method.toUpperCase()}", ${pathTemplate}${optsArg}) as Promise<${returnType}>;`);
             methodLines.push('  }');
             methodLines.push('');
--- a/src/codegen/command-map.ts
+++ b/src/codegen/command-map.ts
@@ -1,6 +1,6 @@
 import type { OpenApiOperation, OpenApiSchema } from './openapi-types';
 import { HTTP_METHODS } from './openapi-types';
-import { normalizeTag, resolveSchemaProps, sanitizeIdentifier } from './util';
+import { normalizeTag, resolveSchemaProps, buildMethodNameMap } from './util';
 
 /**
  * Generate a command map that maps CLI command paths to their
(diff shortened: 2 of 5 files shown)


garretpremo (commented):

All three acceptance criteria from #118 are met and CI is green — the shared `buildMethodNameMap` correctly de-duplicates colliding operationIds across all three emission sites, and generated output for no-collision specs was verified byte-identical to `dev` (petstore, edge-cases, and oas31 fixtures). No blocking issues.

Verified during review:

- The non-null assertions `methodNames.get(...)!` are safe — all three generators guard with `if (!op || !op.operationId) continue;` before any lookup (`client.ts:41`, `commands.ts:33`, `command-map.ts:38`), and `buildMethodNameMap` inserts every op passing the same predicate over the same iteration.
- The determinism/sync-by-construction claim holds: all three generators receive the same `paths` object from `generate()`, so identical iteration order yields identical suffix assignment in each map.
- Tests cover the sharp edges, including the trickiest one — a real `Foo_bar_2` operationId claiming its name first, forcing the collision pair past it via the `used` set.

<details>

- **Test gap: `generateCommands` is never exercised on a collision spec.** The `#118` test block runs only `generateClient` and `generateCommandMap`, but `commands.ts` is the third sync-critical site (it emits `await client.${methodName}(...)`). If a future refactor reverted `commands.ts` to `sanitizeIdentifier`, direct CLI invocation would dispatch to the wrong client method and every test in this PR would still pass. Cheap to close: in the existing collision tests, also run `generateCommands(paths, {})`, assert both `client.Foo_bar(` and `client.Foo_bar_2(` appear, and `expectParses` it.
- Duplicate *identical* raw operationIds (a spec violation) still shadow each other — the `map.has` skip in `buildMethodNameMap` gives both ops the same name. This is pre-existing behavior and inherent to keying by raw id per the issue's suggested design, but a one-sentence JSDoc note would prevent anyone assuming this PR fixed it.
- Command *path* names (`cmdName`, kebab-cased from the raw operationId when verbs collide) have their own theoretical collision class (e.g. case-only-distinct `FooBar`/`fooBar` both kebab to `foo-bar`). Pre-existing and unrelated to sanitization — worth a follow-up issue only if a real spec ever hits it.

</details>

<details>
<summary>Nitpicks</summary>

- The map is rebuilt once inside each of the three generators (3× per `generate()`). O(paths) each, so negligible — only worth hoisting into `generate()` if a fourth emission site appears.
- `for (const [, methods] of Object.entries(paths))` in `buildMethodNameMap` could be `Object.values(paths)`; matching the emitters' `Object.entries` shape is defensible either way.

</details>

PR #123, opened 2026-07-27
feat: make the setup URL default configurable per project (#122)

garretpremo:
Closes #122

## Summary

The interactive setup URL prompt hardcoded `http://localhost:8080` at two sites, with no way for a consumer project to change it. apijack is a generic framework, but every wrapper it generated assumed the wrapped API runs locally on 8080 — so a CLI for an API on any other port handed new users a wrong default on the very first command they ran, and the only workarounds were documenting "ignore the default" or bypassing config with the `<NAME>_URL` env var.

This adds a `defaultUrl` option, settable in `.apijack.json` or passed to `createCli`.

**Scope note:** this is a *prompt default only*. Issue #122's Notes section raised whether `defaultUrl` should also seed the non-interactive resolution chain; the narrower option was chosen deliberately during triage. `src/config.ts` is untouched — `<NAME>_URL` env var → saved config → error behaves exactly as before, including when it errors. Broadening it would change a CLI that fails loudly with "Run `<cli> setup`" into one that silently attempts a request against `defaultUrl` with no credentials.

### New option

`defaultUrl?: string` added to two public interfaces, both optional and additive:

- `ProjectConfig` (`src/project.ts`) — settable in `.apijack.json`
- `CliOptions` (`src/types.ts`) — for consumers with a custom `bin/<cli>.ts` calling `createCli` directly, who may have no `.apijack.json` at all

```json
{
  "specUrl": "/v3/api-docs",
  "generatedDir": ".apijack/generated",
  "defaultUrl": "http://localhost:7080"
}
```

### Plumbing

`.apijack.json` → `loadProjectConfig` → `bin/apijack.ts` → `CliOptions.defaultUrl` → two sinks:

- `cli-builder.ts` → `registerSetupCommand` opts → the `setup` / `login` command
- `cli-builder.ts` → the implicit first-run prompt inside `run()`

This mirrors how the sibling `allowedCidrs` option is threaded through the same call paths. Issue #122's proposed step 4 said "have `cli-builder` read `projectConfig.defaultUrl`" — `cli-builder` never reads project config; `bin/apijack.ts` loads it and passes values into `createCli`, hence the `CliOptions` field.

### Prompt sites

Both sites now derive the prompt label and the prompt's default value from a single variable, so they cannot drift apart:

```ts
const urlDefault = setupUrlDefault(opts?.defaultUrl);
const url = await prompt(`URL [${urlDefault}]: `, urlDefault);
```

`setupUrlDefault` is a small pure helper next to the new exported `DEFAULT_SETUP_URL` constant. It centralises "empty string means unset" and gives a mock-free unit to test, matching this repo's dependency-injection test style. Neither is exported from `src/index.ts` — that file exports nothing from `src/commands/*`, and consumers configure through `CliOptions` / `.apijack.json`.

## Acceptance criteria

- [ ] `defaultUrl` in `.apijack.json` changes the pre-filled URL in both `<cli> setup` and the auto-setup prompt in `cli-builder.ts`
- [ ] `defaultUrl` passed to `createCli` works for consumers with a custom `bin/<cli>.ts` and no `.apijack.json`
- [ ] With `defaultUrl` unset, both prompts still show and accept `http://localhost:8080`
- [ ] `getActiveEnvConfig` / `resolveAuth` behavior is unchanged — no new fallback, no change to when the CLI errors
- [ ] Tests cover set / unset / empty for the resolution helper
- [ ] Public API addition is additive and non-breaking

## Test plan

- [x] `bun test` — 958 pass / 0 fail
- [x] `bunx tsc --noEmit` — only pre-existing TS2578 errors in untouched `src/mcp/**/*.spec.ts`; none in changed files
- [x] Manual smoke, `defaultUrl: "http://localhost:7080"` → `URL [http://localhost:7080]:`
- [x] Manual smoke, unset → `URL [http://localhost:8080]:`
- [x] `grep` confirms `http://localhost:8080` no longer appears in non-test source except `DEFAULT_SETUP_URL` and prose/comments

## Reviewer notes

Two deliberate non-changes, both flagged and accepted during self-review:

- **`src/run-routine.ts:160` threads `allowedCidrs` but not `defaultUrl`.** Cosmetic asymmetry, not a bug — the programmatic path builds its runtime via `_buildRoutineRuntime`, which throws `No active env config` rather than prompting, so the setup prompt is unreachable from `runRoutine`.
- **`src/mcp/tools/config/setup/setup.ts:18`** keeps `http://localhost:8080` in a zod `.describe()` example on a *required* `url` param. The MCP agent always supplies the URL, so there is no default to configure.

**Docs:** there is no in-repo doc file for `.apijack.json` keys (they live on the wiki; `allowedCidrs` is likewise undocumented in-repo). The wiki's *Building a CLI* and *Project Mode* pages need a follow-up edit for `defaultUrl`.

### DIFF
--- a/bin/apijack.ts
+++ b/bin/apijack.ts
@@ -152,6 +152,7 @@ const cli = createCli({
     sessionAuth,
     generatedDir,
     allowedCidrs: projectConfig?.allowedCidrs,
+    defaultUrl: projectConfig?.defaultUrl,
     configPath: join(configDir, 'config.json'),
     customCommandDefaults: projectSettings.customCommands?.defaults,
 });
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -21,7 +21,7 @@ import { registerPluginsCommand } from './commands/plugins/register';
 import { PluginRegistry } from './plugin/registry';
 import { loadPluginPeerInfo, checkPeerRange } from './plugin/peer-version';
 import { PluginPeerMismatchError } from './plugin/errors';
-import { registerSetupCommand, setupAction } from './commands/setup/setup';
+import { registerSetupCommand, setupAction, setupUrlDefault } from './commands/setup/setup';
 import { registerConfigCommand } from './commands/config/register';
 import { registerGenerateCommand } from './commands/generate/generate';
 import { registerUpgradeCommand } from './commands/upgrade/upgrade';
@@ -466,6 +466,7 @@ export function createCli(options: CliOptions): Cli {
                 allowedCidrs: options.allowedCidrs,
                 configPath: options.configPath,
                 displayName,
+                defaultUrl: options.defaultUrl,
             });
             registerConfigCommand(program, cliName, {
                 configPath: options.configPath,
@@ -511,7 +512,8 @@ export function createCli(options: CliOptions): Cli {
             ) {
                 console.log(`${displayName} Setup\n`);
                 const envName = await prompt('Environment name [default]: ', 'default');
-                const url = await prompt('URL [http://localhost:8080]: ', 'http://localhost:8080');
+                const urlDefault = setupUrlDefault(options.defaultUrl);
+                const url = await prompt(`URL [${urlDefault}]: `, urlDefault);
                 const user = await prompt('Username/Email: ');
                 const password = await hiddenPrompt('Password: ');
 
--- a/src/commands/setup/setup.test.ts
+++ b/src/commands/setup/setup.test.ts
@@ -1,5 +1,20 @@
 import { describe, test, expect, mock } from 'bun:test';
-import { setupAction } from './setup';
+import { setupAction, setupUrlDefault, DEFAULT_SETUP_URL } from './setup';
+
+describe('setupUrlDefault', () => {
+    test('falls back to localhost:8080 when no defaultUrl is configured', () => {
+        expect(setupUrlDefault(undefined)).toBe('http://localhost:8080');
+        expect(DEFAULT_SETUP_URL).toBe('http://localhost:8080');
+    });
+
+    test('uses the consumer-configured defaultUrl', () => {
+        expect(setupUrlDefault('http://localhost:7080')).toBe('http://localhost:7080');
+    });
+
+    test('ignores an empty defaultUrl', () => {
+        expect(setupUrlDefault('')).toBe('http://localhost:8080');
+    });
+});
 
 describe('setupAction', () => {
     test('calls saveEnvironment with provided credentials', async () => {
--- a/src/commands/setup/setup.ts
+++ b/src/commands/setup/setup.ts
@@ -3,6 +3,14 @@ import { saveEnvironment, verifyCredentials } from '../../config';
 import type { EnvironmentConfig } from '../../config';
 import { prompt, hiddenPrompt } from '../../prompt';
 
+/** URL offered by the interactive setup prompts when a consumer configures none. */
+export const DEFAULT_SETUP_URL = 'http://localhost:8080';
+
+/** Resolve the URL pre-fill for the setup prompts — the consumer's `defaultUrl`, else the built-in default. */
+export function setupUrlDefault(defaultUrl?: string): string {
+    return defaultUrl || DEFAULT_SETUP_URL;
+}
+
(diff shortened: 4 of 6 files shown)


garretpremo (commented):

All six acceptance criteria are met. The change is additive (three optional fields), the plumbing `.apijack.json` → `loadProjectConfig` → `bin/apijack.ts` → `CliOptions` → both prompt sites is complete and mirrors how `allowedCidrs` is threaded, `src/config.ts` is untouched so the non-interactive resolution chain is unchanged, and CI is green on all 10 checks. Both reviewer-flagged non-changes check out: `src/run-routine.ts` ends at `cli.runRoutine(...)` and never calls `cli.run()`, so the first-run prompt is genuinely unreachable from the programmatic path, and the MCP `setup` tool's `url` is a required param with no default to configure.

<details>

- **No test covers the wiring, only the helper.** `setupUrlDefault` is tested for set/unset/empty, but nothing asserts that a `defaultUrl` actually reaches a prompt. A regression dropping the field at any of the three hops — `bin/apijack.ts:155`, the `registerSetupCommand` opts at `src/cli-builder.ts:469`, or `options.defaultUrl` at `src/cli-builder.ts:515` — would ship green. Both prompt sites import `prompt` directly rather than receiving it injected, so covering them would need a module mock; that constraint is pre-existing and these blocks are untested today, so this isn't something the PR introduced.
- **The prompt block itself is still duplicated.** `src/commands/setup/setup.ts:63` and `src/cli-builder.ts:515` remain near-identical four-prompt sequences ending in a `setupAction` call. The helper removes the risk of the *URL literal* drifting, which is what #122 asked for, but the surrounding block can still diverge. Out of scope for this issue.
- **Wiki follow-up is real and can't ride along.** `wiki/` is gitignored (`.gitignore:11`), so those edits are necessarily a separate push. Two pages enumerate options and now omit `defaultUrl`: the `.apijack.json` field table in `wiki/Project-Mode.md:38-46` (plus the "all fields" example at line 50) and the `CliOptions` list in `wiki/Building-a-CLI.md:104`. Minor correction to the PR body's framing — `allowedCidrs` is undocumented in the *tracked* repo but is documented in both of those wiki pages, so `defaultUrl` warrants the same treatment rather than none.

</details>

<details>
<summary>Nitpicks</summary>

- `setupUrlDefault` uses `||`, so a whitespace-only `defaultUrl` (`" "`) passes through as the pre-fill while `""` correctly falls back. Empty string is the case that matters and it's tested; whitespace is a trivial gap.
- The doc comments on `ProjectConfig.defaultUrl` and `CliOptions.defaultUrl` say "Defaults to `http://localhost:8080`" without mentioning that an empty string is treated as unset — deliberate, tested behavior that's worth one clause.

</details>

garretpremo (commented):

Approved.

- **No test covers the wiring, only the helper.** `setupUrlDefault` is tested for set/unset/empty, but nothing asserts that a `defaultUrl` actually reaches a prompt. A regression dropping the field at any of the three hops — `bin/apijack.ts:155`, the `registerSetupCommand` opts at `src/cli-builder.ts:469`, or `options.defaultUrl` at `src/cli-builder.ts:515` — would ship green. Both prompt sites import `prompt` directly rather than receiving it injected, so covering them would need a module mock; that constraint is pre-existing and these blocks are untested today, so this isn't something the PR introduced.
- **The prompt block itself is still duplicated.** `src/commands/setup/setup.ts:63` and `src/cli-builder.ts:515` remain near-identical four-prompt sequences ending in a `setupAction` call. The helper removes the risk of the *URL literal* drifting, which is what #122 asked for, but the surrounding block can still diverge. Out of scope for this issue.
- **Wiki follow-up is real and can't ride along.** `wiki/` is gitignored (`.gitignore:11`), so those edits are necessarily a separate push. Two pages enumerate options and now omit `defaultUrl`: the `.apijack.json` field table in `wiki/Project-Mode.md:38-46` (plus the "all fields" example at line 50) and the `CliOptions` list in `wiki/Building-a-CLI.md:104`. Minor correction to the PR body's framing — `allowedCidrs` is undocumented in the *tracked* repo but is documented in both of those wiki pages, so `defaultUrl` warrants the same treatment rather than none.

Issue #111, opened 2026-05-09
Wiki: bring Building-a-CLI, Project-Mode, Session-Auth pages up to v1.13.0


garretpremo:
## Background

The public GitHub wiki at https://github.com/normalled/apijack/wiki has drifted from the source since v1.12.0. `README.md` links to it as the primary reference, so first-time users hit stale content immediately. The wiki lives in a separate repo (`github.com/normalled/apijack.wiki`); a local checkout is gitignored at `~/projects/apijack/wiki/`. Updates need a separate `git push` to that repo, not a PR on this one.

This was originally surfaced by a doc-drift audit on 2026-05-04. Three deferred follow-ups from that audit (#98, #99, #100) have all shipped — these wiki gaps are what's left, plus one new gap from v1.13.0 (`dropBaseHeaders`).

### `wiki/Building-a-CLI.md` — `CliOptions` tables

`src/types.ts:18-40` defines three optional fields that don't appear in the "Required fields" / "Optional fields" tables (`Building-a-CLI.md:43-99`):

- `programName?: string` — display branding override (the headline feature of v1.12.0; shown in `--help`, hint strings, error messages)
- `sessionAuth?: SessionAuthConfig` — opt-in cookie/CSRF flow config; cross-reference `wiki/Session-Auth.md`
- `customCommandDefaults?: { requiresAuth?: boolean }` — opt-in default for project-extension commands; cross-reference `wiki/Project-Mode.md`

### `wiki/Project-Mode.md` — `.apijack.json` field table

`src/project.ts:5-13` defines two fields not in the field table (`Project-Mode.md:38-44`):

- `description?: string` — surfaced via `bin/apijack.ts` for branded `--help` in project mode
- `version?: string` — surfaced via `bin/apijack.ts` for `--version` in project mode

Both shipped in #83 (v1.12.0) and are consumed by the shared `apijack` binary.

### `wiki/Session-Auth.md` — missing `dropBaseHeaders`

`SessionAuthConfig.dropBaseHeaders` shipped in v1.13.0 (#107). The wiki has 8 consistent references to `refreshOn` but zero references to `dropBaseHeaders`. The corresponding `CLAUDE.md` section (`"Dropping base-strategy headers post-handshake (opt-in)"`) is the source of truth and can be paraphrased into a parallel subsection on `Session-Auth.md`.

## Acceptance criteria

[...]

PR #126, opened 2026-07-27
chore: promote shipped issues to the 'deployed' label (#125)

garretpremo:
Closes #125

## Summary

Nothing in the release pipeline moved an issue to a terminal label once its code actually reached `main`. `deployed` was applied consistently through #84, then stopped — every issue from #91 onward closed still carrying `ready-for-implement` (a *triage*-state label) and, in most cases, `merged to dev`, whose own description reads "Will be closed once merged to main" and so is stale by definition the moment the release lands.

The 11 affected issues (#91, #98, #99, #100, #106, #109, #111, #115, #116, #118, #122) have been backfilled by hand. This PR is the automation so it stops recurring.

### New `scripts/promote-shipped-issues.sh <release-pr>`

The direct counterpart to final-review's `label-merged-issues.sh`, one stage later in the pipeline — that script applies `merged to dev` at the dev-merge step; this one promotes to `deployed` after the release merges to `main`. It deliberately mirrors the sibling's structure, parse, and best-effort semantics.

Parses the release PR body for closing references, adds `deployed`, strips `ready-for-implement` and `merged to dev`. REST API for the label mutations, matching the existing workaround at `ship.sh:157` (`gh pr edit --add-label` is broken by the projects-classic deprecation).

### Wired into `ship.sh`, not the skill markdown

`ship.sh` gets a `promote_issues` helper called from **both** post-merge paths — after a successful publish, and in the `no publish needed` early-exit branch, which also runs with the PR already merged to `main`. Putting it in `ship.sh` rather than in `ship-release/SKILL.md` means `patch-deployer` inherits the behavior for free, since it shells out to the same script.

It never aborts a release: by the time it runs the code has already landed, so a labelling hiccup must not report a failed ship.

### Docs

Both `ship-release/SKILL.md` and `patch-deployer/SKILL.md` step lists updated (the latter was already stale-by-omission relative to the change).

## Acceptance criteria

- [ ] A release closing issues leaves each one labeled `deployed`, without `ready-for-implement` or `merged to dev`
- [ ] A chore-only release closing no issues is a clean no-op, exit 0
- [ ] Re-running against an already-promoted issue is idempotent
- [ ] A labelling failure does not abort a release that has already published

## Test plan

Verified against live PRs (all are idempotent re-runs against already-shipped releases):

- [x] `promote-shipped-issues.sh 124` → `+ promoted issue #122`, exit 0
- [x] `promote-shipped-issues.sh 121` → promoted #115, #116, #118, exit 0
- [x] `promote-shipped-issues.sh 114` (no closing refs) → friendly message, exit 0
- [x] Missing arg → usage error, exit 1
- [x] Failure accounting → synthetic bogus issue number yields exit 1, so `ship.sh`'s `|| warn` is reachable
- [x] `bash -n` on both scripts

Regex fixture, covering the false-positive case that motivated the leading `\b` — without it, `Discloses #15` would match and the script would strip `ready-for-implement` off an unrelated **open** issue:

```
Closes #10 / Fixed #11 / Resolved #12 / close #13 / Fixes #16   → 10 11 12 13 16   ✓
This prefixes #14 nothing / Discloses #15                       → not matched      ✓
```

## Reviewer notes

Two things worth a second look:

**The `|| true` on the issues pipeline is load-bearing.** Under `set -euo pipefail`, `grep` exiting 1 on no-match aborts the script before the friendly no-op message can run. I hit this as a live regression while testing and added it back with an accurate comment.

**The same bug exists in `label-merged-issues.sh` today** — `.claude/skills/final-review/scripts/label-merged-issues.sh 114` exits 1 silently, making its `if [ -z "$issues" ]` branch dead code. Pre-existing and out of scope here; filed separately.

### DIFF
--- a/.claude/skills/patch-deployer/SKILL.md
+++ b/.claude/skills/patch-deployer/SKILL.md
@@ -121,7 +121,8 @@ The script edits the open `dev → main` PR if one exists, otherwise pushes `dev
 3. Wait for CI
 4. Merge via `gh pr merge --merge --admin`
 5. Watch the publish workflow create the release + publish to npm
-6. Sync main and rebase dev
+6. Promote every issue closed by the release to `deployed`, stripping the now-stale `ready-for-implement` and `merged to dev` (via `scripts/promote-shipped-issues.sh`, driven by the `Closes #N` lines in the PR body)
+7. Sync main and rebase dev
 
 If ship.sh exits non-zero, leave the PR in place for a human and stop.
 
--- a/.claude/skills/ship-release/SKILL.md
+++ b/.claude/skills/ship-release/SKILL.md
@@ -148,7 +148,10 @@ ship.sh will:
 3. Wait for CI
 4. Merge via `gh pr merge --merge --admin`
 5. Watch the publish workflow
-6. Sync main and rebase dev
+6. Promote every issue closed by the release to `deployed`, stripping the now-stale `ready-for-implement` and `merged to dev` (via `scripts/promote-shipped-issues.sh`, driven by the `Closes #N` lines from step 2)
+7. Sync main and rebase dev
+
+Step 6 is why the `Closes #N` lines matter twice over: they auto-close the issues *and* drive the label promotion. A release that closes no issues is a clean no-op. Labelling never aborts a ship — the release is already published by then, so a failure only warns.
 
 If ship.sh fails at any step, it prints what to do. Fix on dev, commit, re-run — it picks up where it left off.
 
--- a/scripts/promote-shipped-issues.sh
+++ b/scripts/promote-shipped-issues.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+# Promote every issue referenced as `Closes #N` / `Fixes #N` / `Resolves #N`
+# in a release PR body to the terminal `deployed` label, and strip the
+# workflow-state labels that are stale once the code reaches main:
+#
+#   ready-for-implement — a triage-state label, meaningless post-ship
+#   merged to dev       — its own description reads "Will be closed once
+#                         merged to main", which has now happened
+#
+# The counterpart to final-review's label-merged-issues.sh, which applies
+# `merged to dev` at the dev-merge step. This runs one stage later, from
+# ship.sh after the release merges to main — so patch-deployer inherits it
+# too, since it shells out to the same ship.sh.
+#
+# Only the curated-PR path carries `Closes #N` lines (ship-release step 2
+# forward-ports them via collect-closes-refs.sh). A release PR that ship.sh
+# auto-generated has none, and is a clean no-op — consistent with GitHub,
+# which wouldn't auto-close those issues either.
+#
+# Idempotent: re-adding an existing label is a no-op for the GitHub API, and
+# deleting an absent one 404s harmlessly.
+#
+# Usage: promote-shipped-issues.sh <release-pr>
+
+set -euo pipefail
+
+source "$(git rev-parse --show-toplevel)/scripts/gh-pin-account.sh"
+
+pr="${1:?release PR number required}"
+repo="normalled/apijack"
+add_label="deployed"
+strip_labels=("ready-for-implement" "merged to dev")
+
+body=$(gh api "repos/$repo/pulls/$pr" --jq '.body // ""')
+
+# Match GitHub's recognized closing keywords:
+#   close / closes / closed
+#   fix / fixes / fixed
+#   resolve / resolves / resolved
+# followed by whitespace, `#`, and a number. Case-insensitive on the keyword.
+# The leading `\b` matters: without it, "discloses #15" would match and this
+# script would strip `ready-for-implement` off an unrelated open issue.
+#
+# `|| true` is load-bearing under `set -euo pipefail`: grep exits 1 when nothing
+# matches, which would abort the script on the legitimate no-closing-refs case
+# before the friendly message below could run.
+issues=$(printf '%s\n' "$body" \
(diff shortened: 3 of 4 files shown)


garretpremo (commented):

All four acceptance criteria from #125 check out, the diff is tightly scoped to the four files the change needs, and the new script is a faithful structural mirror of its sibling `label-merged-issues.sh`. No blocking issues.

**Verification performed:**

- `bash -n scripts/promote-shipped-issues.sh` — clean
- `./scripts/promote-shipped-issues.sh` (no arg) → `release PR number required`, exit 1
- `./scripts/promote-shipped-issues.sh 114` (no closing refs) → friendly message, exit 0 — confirms the `|| true` on the grep pipeline is doing its job under `set -euo pipefail`
- Regex fixture, extended past the one in the PR body with the shapes that actually appear in release PRs:
  ```
  Closes #10 / Fixed #11 / Resolved #12 / close #13 / Fixes #16   → 10 11 12 13 16  ✓
  Discloses #15 / This prefixes #14 / see PR (#17) / issue #18    → not matched     ✓
  fix: something (#19) / chore: ... (#125)                        → not matched     ✓
  ```
  That last pair matters: it is exactly what `ship.sh`'s auto-generated `## Commits` body looks like, and confirms the auto-generated path is a genuine no-op rather than a source of false promotions. PR #124's real body (`Configurable setup URL default (#123)` in the features list) likewise yields only `122`.
- Label-path encoding: `gh api repos/normalled/apijack/labels/merged%20to%20dev` resolves to the real label, so `gh` passes `%20` through the path rather than re-encoding the `%`. The `${label// /%20}` substitution in the DELETE is sound.
- `deployed` and `merged to dev` both exist as repo labels.
- `PR_NUM` is assigned at `ship.sh:164`, ahead of both `promote_issues` call sites (240 and 263) — no unbound-variable hazard under `set -u`.
- The `|| warn` wrapper returns 0, so criterion 4 holds: a labelling failure cannot abort an already-published release.
- Follow-up issue #127 for the sibling's missing `|| true` is filed as claimed.

Wiring into `ship.sh` rather than the skill markdown is the right call — `patch-deployer` inheriting it for free is worth more than the marginal legibility of putting it in the SKILL step list.

<details>

- **Comment undersells the exit semantics.** The comment above `[ "$failed" -eq 0 ]` reads "Exit non-zero when nothing could be promoted," but the code exits non-zero when *any* issue fails, even if others succeeded. The code's behavior is the better of the two — a partial failure shouldn't be swallowed — so this is a comment fix, not a code fix.

- **The colon form of the closing keyword isn't matched.** GitHub accepts `Closes: #122`, but `[[:space:]]+` requires whitespace immediately after the keyword, so that line is silently skipped. Unreachable through the normal pipeline, since `collect-closes-refs.sh` emits colon-less `Closes #N` lines — but release PR bodies are hand-curated in ship-release step 3, and a human typing the colon form would get a silently unpromoted issue. Same gap for `owner/repo#N` and full-URL references. Inherited verbatim from `label-merged-issues.sh`, so fixing it is a two-file change; reasonable to defer alongside #127.

- **CI on this PR only ran the auto-labeler.** `ci.yml` and `e2e.yml` both carry `paths:` filters covering `src/**`, `tests/**`, and build config — `scripts/**` and `.claude/**` match none of them. That's correct behavior for a change that touches no testable surface area, and the branch heads stay verified by the push-triggered runs, but it does mean this PR has no green test signal beyond the manual verification above. Noting it so the absence isn't mistaken for a stuck check.

- **No automated coverage for the script.** Consistent with the rest of `scripts/` (none of which is tested), and the regex is the only part with real logic in it. If the parse block ever grows a third copy, a fixture-driven test over the extraction pipeline would be cheap insurance.

</details>

<details>
<summary>Nitpicks</summary>

- `promote_issues` re-runs `git rev-parse --show-toplevel` on each call; `ship.sh:4` already resolves the same path when sourcing `gh-pin-account.sh`. Hoisting it to a variable at the top would save the subprocess and read a little cleaner.

- The `## Acceptance criteria` checkboxes in the PR body are still unticked while the `## Test plan` ones are ticked. Cosmetic, and the test plan clearly covers all four.

</details>

garretpremo (commented):

Approved.

- **Comment undersells the exit semantics.** The comment above `[ "$failed" -eq 0 ]` reads "Exit non-zero when nothing could be promoted," but the code exits non-zero when *any* issue fails, even if others succeeded. The code's behavior is the better of the two — a partial failure shouldn't be swallowed — so this is a comment fix, not a code fix.

- **The colon form of the closing keyword isn't matched.** GitHub accepts `Closes: #122`, but `[[:space:]]+` requires whitespace immediately after the keyword, so that line is silently skipped. Unreachable through the normal pipeline, since `collect-closes-refs.sh` emits colon-less `Closes #N` lines — but release PR bodies are hand-curated in ship-release step 3, and a human typing the colon form would get a silently unpromoted issue. Same gap for `owner/repo#N` and full-URL references. Inherited verbatim from `label-merged-issues.sh`, so fixing it is a two-file change; reasonable to defer alongside #127.

- **CI on this PR only ran the auto-labeler.** `ci.yml` and `e2e.yml` both carry `paths:` filters covering `src/**`, `tests/**`, and build config — `scripts/**` and `.claude/**` match none of them. That's correct behavior for a change that touches no testable surface area, and the branch heads stay verified by the push-triggered runs, but it does mean this PR has no green test signal beyond the manual verification above. Noting it so the absence isn't mistaken for a stuck check.

- **No automated coverage for the script.** Consistent with the rest of `scripts/` (none of which is tested), and the regex is the only part with real logic in it. If the parse block ever grows a third copy, a fixture-driven test over the extraction pipeline would be cheap insurance.

PR #80, opened 2026-04-28
chore: forward-port Closes #N refs into release PR bodies

garretpremo:
## Summary

Fixes a workflow gap surfaced by v1.10.2: six issues (#65–#70) all had `Closes #N` in their issue PR bodies, all merged cleanly into `dev`, but none auto-closed when PR #78 (dev → main) merged. GitHub only honors closing keywords on PRs targeting the default branch (`main`), so closures on dev-targeted PRs are never seen.

The fix is to forward-port those keywords into the release PR body. New `scripts/collect-closes-refs.sh` walks every PR referenced in the release range, scrapes their bodies for `Closes/Fixes/Resolves #N`, and emits deduped `Closes #N` lines for the release PR template to consume.

## What changes

- **`scripts/collect-closes-refs.sh`** — reads `/tmp/apijack-ship-commits.txt` (or `$1`), runs `gh pr view` per referenced PR, prints sorted unique `Closes #N` lines. Issue numbers harmlessly fail `gh pr view` and are skipped.
- **`ship-release` SKILL.md** — adds a "Collect closing references" sub-step in step 2 and inserts `Closes #N` lines at the top of the body template in step 3.
- **`patch-deployer` SKILL.md** — same treatment as a new step 3a, with closing keywords at the top of the body template in step 4.

## Acceptance criteria

- [x] Running the script against v1.10.2's commit list produces exactly `Closes #65` through `Closes #70`.
- [x] Empty output is fine for chore-only releases that resolved no tracked issues — both skills note this.
- [x] Script skips silently when an extracted `#NN` is an issue (not a PR).
- [x] Both skill body templates lead with `Closes #N` lines so GitHub will see them on the dev → main merge.

## Test plan

- [x] Verified end-to-end against v1.10.2 commit data: script outputs the expected 6 lines.
- [x] `bash -n scripts/collect-closes-refs.sh` clean.
- [ ] Will exercise on the next release shipped via `ship-release` or `patch-deployer`.

PR #130, opened 2026-07-28
fix: ignore closing refs inside code when labeling issues (#129, #127)

garretpremo:
Closes #129
Closes #127

## Summary

Three scripts extracted `Closes #N`-style references from PR bodies with a plain `grep`, which has no notion of markdown structure. A closing keyword written inside a fenced code block or an inline code span matched exactly as if it were a real reference — and the scripts then mutated labels on whatever issue number followed.

This is not theoretical. PR #126's body contained a regex test fixture inside a fenced block; running `label-merged-issues.sh 126` parsed it as five real references and applied `merged to dev` to PRs #10, #11, #12, #13 and #16 — unrelated release PRs from months earlier, reverted by hand. The blast radius is worse for `promote-shipped-issues.sh`, which doesn't only add a label but **strips** `ready-for-implement`, silently destroying the triage state of an open, unshipped issue.

Release PR bodies are where this recurs most easily: `ship-release` and `patch-deployer` both instruct the author to paste representative code and YAML into the body, and the skills' own docs quote the closing-keyword convention.

This PR replaces all three regexes with one shared, network-free extractor that strips code before matching. It also picks up #127 — the missing `|| true` in `label-merged-issues.sh` — because the fix rewrites those exact lines, and putting the guard inside the extractor fixes it once for every caller.

### New: `scripts/extract-closing-refs.sh`

Takes a markdown body on stdin or `--body-file`, prints issue numbers one per line, deduped and ascending. Network-free by design — callers keep their own fetch, which legitimately differs between them (the label scripts hit `gh api repos/.../pulls/N`; `collect-closes-refs.sh` uses `gh pr view` so it can skip numbers that turn out to be issues rather than PRs).

Stripping, via awk, before the keyword match:

- **Fenced blocks** — ``` and `~~~`, per CommonMark: 3+ delimiters, up to 3 spaces of indent, closing fence at least as long as the opening one, backtick info strings that can't themselves contain a backtick.
- **Inline code spans** — scanned by matching delimiter-run *length*, so ``` ``Closes #10`` ``` is removed whole. The naive `` `[^`]*` `` strip would consume just the two delimiter pairs and leave `Closes #10` behind as a live reference.

Three gaps are deliberate and documented in the script header, each pinned by a test. All of them leave a reference **visible** (a possible false positive) rather than hide a real one — the safer direction:

- indented code blocks and blockquotes aren't stripped (a blockquote can legitimately carry a real reference)
- a fence indented 4+ spaces isn't recognized, so a fence nested in a list item leaks. Allowing any indent would be worse: an indented-code-block fence marker would open a phantom fence and swallow every real reference after it.
- code spans are scanned per line, so a span wrapping across a newline leaks

The word boundary is spelled `(^|[^[:alnum:]_])` rather than `\b`: GNU grep honors `\b` in ERE but BSD/macOS grep uses `[[:<:]]`, where `\b` degrades to a literal `b` and matches nothing. `collect-closes-refs.sh` would have inherited that hazard newly. The awk program avoids interval expressions for the same reason and is verified identical under gawk and mawk.

### Modified: the three call sites

| Script | Before | After |
|---|---|---|
| `.claude/skills/final-review/scripts/label-merged-issues.sh` | own regex, **no** `|| true` | shared extractor |
| `scripts/promote-shipped-issues.sh` | own regex, `|| true` | shared extractor |
| `scripts/collect-closes-refs.sh` | **drifted** regex, `|| true` | shared extractor |

`collect-closes-refs.sh` is the upstream vector: whatever it pulls out of an issue-PR body gets written into the release PR body as `Closes #N`, which `promote-shipped-issues.sh` then acts on. A keyword quoted in a fenced example there becomes a real label mutation two steps later. Adopting the shared implementation also settles its drift — it previously matched only `Closes|Fixes|Resolves` (missing the bare `close`/`fixed`/`resolved` forms GitHub honors) and lacked the leading `\b` that rejects `Discloses #15`.

Measured against PR #126's real body (committed as `tests/fixtures/pr-126-body.md`):

```
OLD label-script regex:        10 11 12 13 16 125
OLD collect-closes-refs regex: 10 14 15 16 125      ← 14 and 15 from "prefixes #14" / "Discloses #15"
NEW shared extractor:          125
```

Sweeping the old regex against the new extractor over every PR body in the repo, exactly two differ — and both differ in the correct direction. Besides #126, review turned up a second, previously unnoticed false positive on **PR #80** (`old=[65 70] → new=[]`), whose body quotes two refs inside inline code spans. **Zero false negatives across every real body** — that's the dangerous direction, and it's clean.

### Error propagation

Empty-output-exit-0 is the extractor's contract for "this PR closes nothing", so an internal failure must never produce the same signal — otherwise a broken extractor reads as a chore-only release and every issue in it silently never auto-closes. Two guards:

- awk runs **outside** the pipeline, so `set -e` catches it; the trailing tolerance is scoped to grep's exit 1 (`|| [ $? -eq 1 ]`) rather than a blanket `|| true`.
- `collect-closes-refs.sh` checks the extractor is executable up front. It's the one call site where failure would otherwise be silent — the extractor runs inside a process substitution whose status is never examined, inside a loop that `continue`s freely. The other two assign from a command substitution and already abort under `pipefail`.

This PR body is its own regression test, for the same reason #126's was a live grenade — it quotes the closing-keyword convention throughout:

```
OLD label-script regex over this body: 10 125 127 129
NEW shared extractor:                  127 129
```

### #127: no closing refs is not a failure

`label-merged-issues.sh` ran its pipeline under `set -euo pipefail` with no `|| true`, so grep's exit 1 on no-match aborted the script and made its own `if [ -z "$issues" ]` friendly branch dead code. The `|| true` now lives inside the extractor, so all three callers inherit it:

```
$ .claude/skills/final-review/scripts/label-merged-issues.sh 114
label-merged-issues: no closing references in PR #114 body; nothing to label.
$ echo $?
0
```

## Acceptance criteria

- [ ] A closing keyword inside a fenced code block is ignored by all three scripts
- [ ] A closing keyword inside an inline code span is ignored, including multi-backtick delimiters
- [ ] Genuine references outside code are still matched, across all of GitHub's keyword forms (`close`/`closes`/`closed`, `fix`/`fixes`/`fixed`, `resolve`/`resolves`/`resolved`)
- [ ] `Discloses #15`-style false positives stay rejected (the leading `\b` is preserved)
- [ ] PR #126's body through the extractor yields only `125`, and none of 10/11/12/13/16
- [ ] All three scripts share one extraction implementation
- [ ] `collect-closes-refs.sh` adopts the unified keyword set
- [ ] `label-merged-issues.sh <pr-with-no-closing-refs>` prints its message and exits 0 (#127)
- [ ] Normal labelling behavior is unchanged for PRs that do have closing refs (#127)

## Test plan

- [x] `bun test` — 983 pass / 0 fail (25 new in `tests/extract-closing-refs.test.ts`)
- [x] `bash -n` on all four scripts
- [x] `label-merged-issues.sh 114` (no closing refs) → friendly message, exit 0 — the #127 repro
- [x] `promote-shipped-issues.sh 114` (no closing refs) → friendly message, exit 0
- [x] `collect-closes-refs.sh` over a commit range containing #126 → `Closes #125` only
- [x] `extract-closing-refs.sh --body-file tests/fixtures/pr-126-body.md` → `125`
- [x] Swept old regex vs. new extractor over every PR body in the repo → only #126 and #80 differ, both false positives removed, zero false negatives
- [x] awk shimmed to fail → extractor exits non-zero rather than reporting "no refs"
- [x] extractor chmod -x → `collect-closes-refs.sh` exits 1 with a message (was a silent exit 0)
- [x] awk program byte-identical under gawk 5.2 and mawk 1.3.4

New test coverage: every keyword form; dedupe and numeric sort; mid-sentence and multiple refs per line; `Discloses`/`prefixes`/`affixes` rejection; backtick and tilde fences; info strings; longer closing fences; 3-space fence indent; single- and multi-backtick inline spans; unmatched backtick as literal; unterminated fence (result and stderr warning); CRLF bodies; the three documented gaps; the #126 regression; empty-result exit-0 cases; all three usage forms; awk-failure propagation.

## Out of scope

Deferred to a follow-up, per triage on #129:

- GitHub's `closingIssuesReferences` GraphQL field as a cross-check in `promote-shipped-issues.sh`. It's fence-aware and authoritative, but only reports references closing against the **default** branch — so it can't serve `label-merged-issues.sh`, whose PRs target `dev`. Adding a second source of truth for one script but not the others reintroduces the drift this PR removes.
- Stripping indented code blocks and blockquotes.

### DIFF
--- a/.claude/skills/final-review/scripts/label-merged-issues.sh
+++ b/.claude/skills/final-review/scripts/label-merged-issues.sh
@@ -19,7 +19,10 @@
 
 set -euo pipefail
 
-source "$(git rev-parse --show-toplevel)/scripts/gh-pin-account.sh"
+repo_root="$(git rev-parse --show-toplevel)"
+source "$repo_root/scripts/gh-pin-account.sh"
+
+extract="$repo_root/scripts/extract-closing-refs.sh"
 
 pr="${1:?pr number required}"
 repo="normalled/apijack"
@@ -27,15 +30,16 @@ label="merged to dev"
 
 body=$(gh api "repos/$repo/pulls/$pr" --jq '.body // ""')
 
-# Match GitHub's recognized closing keywords:
-#   close / closes / closed
-#   fix / fixes / fixed
-#   resolve / resolves / resolved
-# followed by whitespace, `#`, and a number. Case-insensitive on the keyword.
-issues=$(printf '%s\n' "$body" \
-    | grep -oiE '\b(close[sd]?|fix(es|ed)?|resolve[sd]?)[[:space:]]+#[0-9]+' \
-    | grep -oE '[0-9]+' \
-    | sort -un)
+# Keyword matching lives in scripts/extract-closing-refs.sh, shared with
+# promote-shipped-issues.sh and collect-closes-refs.sh. It strips fenced code
+# blocks and inline code spans first, so a body that merely documents the
+# `Closes #N` convention doesn't get read as a reference — the failure that
+# labeled five unrelated PRs during #126.
+#
+# It exits 0 with no output when there are no references, which keeps the
+# friendly branch below reachable under `set -euo pipefail` (#127) — and
+# non-zero on an internal failure, which aborts here rather than labeling nothing.
+issues=$(printf '%s\n' "$body" | "$extract" -)
 
 if [ -z "$issues" ]; then
     echo "label-merged-issues: no closing references in PR #$pr body; nothing to label."
--- a/scripts/collect-closes-refs.sh
+++ b/scripts/collect-closes-refs.sh
@@ -30,14 +30,39 @@ fi
 # no referenced PRs is a legitimate empty-output case, not a failure.
 PR_NUMS=$(grep -oE '#[0-9]+' "$COMMITS_FILE" | sed 's/#//' | sort -un || true)
 
+# Resolved relative to this script rather than via `git rev-parse`, so the
+# default /tmp commits-file usage still works from outside a checkout.
+EXTRACT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/extract-closing-refs.sh"
+
+# The extractor runs inside a process substitution below, whose exit status is
+# never examined, inside a loop that `continue`s freely — so unlike the other
+# two call sites, a missing or broken extractor here would fail SILENTLY and
+# yield an empty ref list. That is indistinguishable from a chore-only release,
+# and the release PR body would ship with no `Closes #N` lines at all, leaving
+# every issue in the release to never auto-close. Fail loudly instead.
+if [ ! -x "$EXTRACT" ]; then
+    echo "collect-closes-refs: extractor missing or not executable: $EXTRACT" >&2
+    exit 1
+fi
+
 declare -A SEEN
 for pr in $PR_NUMS; do
     BODY=$(gh pr view "$pr" --json body --jq '.body' 2>/dev/null) || continue
     [ -z "$BODY" ] && continue
+    # Shared with label-merged-issues.sh and promote-shipped-issues.sh. This
+    # call site is the upstream vector for #129: whatever it pulls out of an
+    # issue PR body gets written into the release PR body as `Closes #N`, which
+    # promote-shipped-issues.sh then acts on. A keyword quoted in a fenced
+    # example here becomes a real label mutation two steps later.
+    #
+    # Adopting the shared implementation also settles the drift this script had
+    # against the other two: it previously matched only Closes/Fixes/Resolves
+    # (missing the bare `close`/`fixed`/`resolved` forms GitHub honors) and
+    # lacked the leading `\b` that rejects `Discloses #15`.
     while IFS= read -r num; do
         [ -z "$num" ] && continue
         SEEN["$num"]=1
-    done < <(echo "$BODY" | grep -ioE '(Closes|Fixes|Resolves)[[:space:]]+#[0-9]+' | grep -oE '[0-9]+' || true)
+    done < <(printf '%s\n' "$BODY" | "$EXTRACT" -)
 done
 
 for num in "${!SEEN[@]}"; do
--- a/scripts/extract-closing-refs.sh
+++ b/scripts/extract-closing-refs.sh
@@ -28,9 +28,9 @@
 #   - indented code blocks and blockquotes are not stripped (per #129: a
 #     blockquote can legitimately carry a real reference)
 #   - a fence indented 4+ spaces is not recognized, so a fence nested inside a
-#     list item leaks its contents. Allowing any indent would be worse: an
-#     indented-code-block fence marker would open a phantom fence and swallow
-#     every real reference after it.
+#     list item leaks its contents (#134). Allowing any indent would be worse:
+#     an indented-code-block fence marker would open a phantom fence and
+#     swallow every real reference after it.
 #   - code spans are scanned per line, so a span that wraps across a newline
 #     leaks its contents
 #
--- a/scripts/extract-closing-refs.sh
+++ b/scripts/extract-closing-refs.sh
@@ -1,0 +1,182 @@
+#!/usr/bin/env bash
+# extract-closing-refs.sh
+#
+# Read a markdown body (PR or issue) and print the issue number of every
+# GitHub closing reference it contains — one per line, deduped, ascending.
+#
+# The single extraction implementation shared by:
+#   .claude/skills/final-review/scripts/label-merged-issues.sh
+#   scripts/promote-shipped-issues.sh
+#   scripts/collect-closes-refs.sh
+#
+# Each of those used to carry its own copy of the regex, which drifted apart
+# and — more seriously — matched closing keywords written inside code. A PR
+# body that documents the `Closes #N` convention, or pastes a regex fixture
+# into a fenced block, is not making five closing references. Before #129 the
+# scripts read it as if it were: running label-merged-issues.sh against #126,
+# whose body contained exactly such a fixture, applied `merged to dev` to five
+# unrelated release PRs from months earlier.
+#
+# So: strip code before matching.
+#   - fenced blocks (``` and ~~~, per CommonMark: 3+ delimiters, up to 3
+#     spaces of indent, closing fence at least as long as the opening one)
+#   - inline code spans (`x`, ``x``, ...), scanned by matching delimiter-run
+#     length so ``Closes #10`` isn't half-stripped into a live reference
+#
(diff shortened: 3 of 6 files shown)


garretpremo (commented):

Every acceptance criterion on #129 and #127 checks out, and I reproduced the PR's central claims independently rather than taking the test plan on faith. The extractor is the right shape for the bug: it's network-free, the three call sites now share one implementation with no copies left elsewhere in the repo, and each documented gap errs toward a visible false positive rather than a hidden real reference — the safe direction, and each one is pinned by a test.

**Verified independently:**

- Re-ran the old-regex vs. new-extractor sweep over every PR body in the repo (all states, 200-limit). Exactly three differ: #126 (`old=[10 11 12 13 16 125] → new=[125]`), #80 (`old=[65 70] → new=[]`), and this PR's own body (`old=[10 125 127 129] → new=[127 129]`). All three differences remove false positives. **Zero false negatives** — the dangerous direction is clean. #80's refs turned out to be a test-plan sentence describing script output, correctly dropped.
- `extract-closing-refs.sh --body-file tests/fixtures/pr-126-body.md` → `125`, exit 0.
- #127 repro against live PRs: `label-merged-issues.sh 114` and `promote-shipped-issues.sh 114` both print the friendly message and exit 0.
- `collect-closes-refs.sh` over a commit range → correct `Closes #N` lines; with the extractor `chmod -x`, it exits 1 with a message instead of silently empty.
- The 25 new tests pass under **gawk 5, mawk, and busybox awk** (25/25 each), which backs the no-interval-expressions portability argument.
- Full suite 982/983 in a scratch worktree; the single failure is `getPluginPaths() > sourceDir points to project root`, path-dependent on the worktree location and unrelated. CI on the branch is green across all 9 checks.
- `bash -n` clean on all four scripts. `scripts/` is not in `package.json` `files`, so nothing here reaches the published package.

Choosing `(^|[^[:alnum:]_])` over `\b` is the right call and the reasoning holds — `collect-closes-refs.sh` would have newly inherited the BSD-grep hazard. Scope is tight: four scripts, one test file, one fixture, no unrelated changes, no public API surface touched.

<details>

- **One narrow path where an internal failure could still read as "no references."** The header states that empty-output-exit-0 must never be produced by an internal error, and the awk stage is correctly hoisted out of the pipeline to guarantee that. The final pipeline is a partial exception: under `pipefail`, bash reports the *rightmost* non-zero status, so if the first `grep` errored with exit 2 the second `grep` would exit 1 on empty input and the pipeline would report `1` — which `|| [ $? -eq 1 ]` then tolerates. Confirmed empirically: `set -o pipefail; (sh -c 'exit 2' | grep -oE '[0-9]+' | sort -un)` reports status 1. In practice this is unreachable — the regex is a static constant and the input is a pipe, so grep exit 2 needs an I/O error — and the stage that can realistically fail (awk) is already handled correctly. Noting it because the header makes a stronger absolute claim than the code delivers, not because it needs fixing here.
- **`collect-closes-refs.sh` still has a silent runtime path.** The up-front `-x` check closes the missing/non-executable case, which is the likely one. A *runtime* extractor failure inside the process substitution is still unexamined and yields an empty ref list — the same silent-empty outcome the comment above it warns about. Not a regression (the old code carried a blanket `|| true`), and the comment is accurate about why this call site is the risky one.
- **The colon form isn't matched.** `Closes: #7` yields nothing, while `Closes #7` matches. I believe GitHub's linked-issue parser accepts the colon variant, so this may be a real (if rare) false negative. It's pre-existing — all three old regexes required `[[:space:]]+` directly after the keyword — so nothing regresses, and the win is that the extractor is now the single place to change it if you decide it matters. Worth confirming against GitHub's actual behavior before acting.

</details>

<details>
<summary>Nitpicks</summary>

- The three call sites resolve the extractor two different ways: `label-merged-issues.sh` via `git rev-parse --show-toplevel`, the other two via `${BASH_SOURCE[0]}`. Both are justified in context — `collect-closes-refs.sh` documents why it can't use `git rev-parse`, and `label-merged-issues.sh` was already using `git rev-parse` for `gh-pin-account.sh` — but a reader hitting them in sequence may wonder whether the difference is deliberate.
- The `#134` gap is tracked and open, so the deferred fence-indent case isn't a dangling reference. Confirmed.

</details>

garretpremo (commented):

Approved.

- **One narrow path where an internal failure could still read as "no references."** The header states that empty-output-exit-0 must never be produced by an internal error, and the awk stage is correctly hoisted out of the pipeline to guarantee that. The final pipeline is a partial exception: under `pipefail`, bash reports the *rightmost* non-zero status, so if the first `grep` errored with exit 2 the second `grep` would exit 1 on empty input and the pipeline would report `1` — which `|| [ $? -eq 1 ]` then tolerates. Confirmed empirically: `set -o pipefail; (sh -c 'exit 2' | grep -oE '[0-9]+' | sort -un)` reports status 1. In practice this is unreachable — the regex is a static constant and the input is a pipe, so grep exit 2 needs an I/O error — and the stage that can realistically fail (awk) is already handled correctly. Noting it because the header makes a stronger absolute claim than the code delivers, not because it needs fixing here.
- **`collect-closes-refs.sh` still has a silent runtime path.** The up-front `-x` check closes the missing/non-executable case, which is the likely one. A *runtime* extractor failure inside the process substitution is still unexamined and yields an empty ref list — the same silent-empty outcome the comment above it warns about. Not a regression (the old code carried a blanket `|| true`), and the comment is accurate about why this call site is the risky one.
- **The colon form isn't matched.** `Closes: #7` yields nothing, while `Closes #7` matches. GitHub's linked-issue parser is believed to accept the colon variant, so this may be a real (if rare) false negative. It's pre-existing — all three old regexes required `[[:space:]]+` directly after the keyword — so nothing regresses, and the win is that the extractor is now the single place to change it if you decide it matters. Worth confirming against GitHub's actual behavior before acting.

### Note for a human: post-merge issue labeling skipped

Step 7 of this workflow runs `label-merged-issues.sh`, which reads the PR body and applies `merged to dev` to each referenced issue. The copy of that script in the runner's checkout predates this PR and still carries the old `grep -oiE '\b(close[sd]?|...)'` regex — the exact bug being fixed here. Against this PR's own body it extracts `10 125 127 129`, so running it would apply `merged to dev` to **#10 and #125**, which this PR neither closes nor touches.

I skipped that step rather than reproduce the bug during its own fix. **#127 and #129 need `merged to dev` applied by hand** (or by re-running `label-merged-issues.sh 130` from a checkout that includes this merge, where the shared extractor correctly yields `127 129`). Everything else — approval, label transition, and the SHA-pinned merge — completed normally.

PR #131, opened 2026-07-28
ci: trigger stable publishes on a tag push, not a commit-message match (#128)

garretpremo:
Closes #128

## Summary

`publish.yml` gated stable npm publishes by string-matching commit messages: `contains(toJSON(github.event.commits.*.message), 'chore(release):')` scanned **every** commit in a push to `main`. Any commit whose *text* happened to contain that literal armed a publish — a revert (`Revert "chore(release): v1.15.0"`), a docs or skill commit quoting the release convention (this repo's own `ship-release/SKILL.md` documents that exact string), or a branch merged to `main` carrying an older bump commit. The only backstop was npm refusing to republish an existing version, which fails open if `package.json` were also bumped.

Release intent now lives out-of-band in a git tag — you cannot accidentally write a git tag in prose. `ship.sh` tags the merge commit after the release PR merges and pushes it; that tag push is the sole publish trigger. Merging to `main` is now completely inert.

### `.github/workflows/publish.yml`

Trigger moves off `main` and onto tags:

```yaml
on:
  push:
    branches: [next]
    tags: ['v*']

jobs:
  publish:
    if: startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-')
```

- The redundant `Tag release` step (old `publish.yml:41-48`) is deleted — it derived the tag from `package.json` *after* the fact.
- `gh release create` became idempotent (`gh release view … || create`, else `edit`). Previously, re-running the workflow after an npm publish flake died on "release already exists" *before* reaching `npm publish`.
- Release notes are unchanged: `github.sha` on a tag push is the **commit** sha, so `commits/${{ github.sha }}/pulls` still resolves the tagged merge commit to the dev → main release PR.

### `scripts/ship.sh`

- Resolves the merge commit via `gh pr view --json mergeCommit` **with retries** — `gh pr merge` returns when the REST call completes, but `gh pr view` reads GraphQL, which is eventually consistent and briefly returns null.
- Refuses to push a tag that already exists on `origin` (the push would be a silent no-op and no workflow would fire) or that points at a commit other than the release commit.
- **New resume path.** The merge → tag window is the one place a failure now leaves a release merged-and-bumped but unpublished — and because `dev` then has nothing ahead of `main`, a plain re-run would dead-end at the preflight's "No commits ahead of main." `resume_untagged_release()` detects that exact state (main's version has no corresponding tag, a merged dev→main PR exists) and finishes the release. It returns silently in every other state.
- Dropped the 12×5s poll for a run that a branch-push trigger may never have started, along with its exit-0 "no publish needed" path. Run discovery now filters by `--commit "$MERGE_SHA"` (matching the existing `wait-for-publish-next.sh` pattern) and fails loudly.

### Docs

`ship-release/SKILL.md` and `patch-deployer/SKILL.md` both enumerate ship.sh's steps; added the tag step and renumbered. Corrected the "picks up where it left off" claim to describe the post-merge window explicitly.

## Note on the tag glob

The issue proposed `tags: ['v[0-9]+.[0-9]+.[0-9]+']`. I kept `v*` and moved the prerelease exclusion into the `if:`. GitHub documents `+` support in ref filters, but I could not verify `[]` range support or whether patterns anchor to the full ref — and the failure modes are asymmetric. A glob too *loose* lets a prerelease through, which the `if:` catches; a glob too *strict* means a stable release silently never publishes and you find out at ship time. `contains`/`startsWith` semantics are unambiguous, so the precision lives there.

This matters because `publish-next` creates prerelease tags: `gh release create "v1.16.0-next.0"` creates a git tag matching `v*`. Today those don't trigger workflows only because they're created with `GITHUB_TOKEN`. The `!contains(github.ref, '-')` gate makes that explicit rather than a dependency on that behaviour holding. (`github.ref` is `refs/tags/<tag>` — it never contains `owner/repo`, so `normalled/apijack`'s hyphen is irrelevant.)

## Acceptance criteria

- [ ] Merging anything to `main` without pushing a tag never publishes to npm
- [ ] Pushing a `vX.Y.Z` tag publishes that commit to npm and creates the GitHub release
- [ ] The GitHub release body is still the curated release-PR body, not a commit dump
- [ ] `next` prerelease publishing is unaffected
- [ ] Prerelease tags (`v1.16.0-next.0`) cannot arm the stable publish job
- [ ] `ship.sh` tags the merge commit on `main` and no longer relies on polling for a run that may not exist
- [ ] The redundant `Tag release` workflow step is removed

## Test plan

- [x] `bun test` — 958 pass / 0 fail
- [x] `bash -n scripts/ship.sh`; `publish.yml` parses and the trigger/`if:` fields resolve to the intended scalars
- [x] New ship.sh helpers exercised with stubbed `git`/`gh`:
  - tag already on `origin` → refuses instead of pushing a no-op
  - local tag at the wrong commit → refuses with a `git tag -d` hint
  - `mergeCommit` null for 2 polls then populated → resolves on attempt 3
  - `resume_untagged_release` across 4 states: tag exists locally / on origin / no merged PR → all return silently; merged-bumped-untagged → tags, promotes, syncs
- [x] Verified `set -e` does not trip on the `X && return 0` guards in the resume path

**Not verifiable pre-merge:** the tag trigger itself only exercises on a real release. The first ship after this merges is the live test — if no publish run appears, the tag is already pushed and `publish.yml` can be re-run from the Actions tab.

### DIFF
--- a/.claude/skills/patch-deployer/SKILL.md
+++ b/.claude/skills/patch-deployer/SKILL.md
@@ -120,9 +120,10 @@ The script edits the open `dev → main` PR if one exists, otherwise pushes `dev
 2. Find the existing PR (won't overwrite the curated title/body)
 3. Wait for CI
 4. Merge via `gh pr merge --merge --admin`
-5. Watch the publish workflow create the release + publish to npm
-6. Promote every issue closed by the release to `deployed`, stripping the now-stale `ready-for-implement` and `merged to dev` (via `scripts/promote-shipped-issues.sh`, driven by the `Closes #N` lines in the PR body)
-7. Sync main and rebase dev
+5. Tag the merge commit `vX.Y.Z` and push the tag — the tag push is what triggers `publish.yml`; merging alone never publishes
+6. Watch the publish workflow create the release + publish to npm
+7. Promote every issue closed by the release to `deployed`, stripping the now-stale `ready-for-implement` and `merged to dev` (via `scripts/promote-shipped-issues.sh`, driven by the `Closes #N` lines in the PR body)
+8. Sync main and rebase dev
 
 If ship.sh exits non-zero, leave the PR in place for a human and stop.
 
--- a/.claude/skills/ship-release/SKILL.md
+++ b/.claude/skills/ship-release/SKILL.md
@@ -147,13 +147,14 @@ ship.sh will:
 2. Find the existing PR (won't overwrite the curated title/body)
 3. Wait for CI
 4. Merge via `gh pr merge --merge --admin`
-5. Watch the publish workflow
-6. Promote every issue closed by the release to `deployed`, stripping the now-stale `ready-for-implement` and `merged to dev` (via `scripts/promote-shipped-issues.sh`, driven by the `Closes #N` lines from step 2)
-7. Sync main and rebase dev
+5. Tag the merge commit `vX.Y.Z` and push the tag — the tag push is what triggers `publish.yml`; merging alone never publishes
+6. Watch the publish workflow
+7. Promote every issue closed by the release to `deployed`, stripping the now-stale `ready-for-implement` and `merged to dev` (via `scripts/promote-shipped-issues.sh`, driven by the `Closes #N` lines from step 2)
+8. Sync main and rebase dev
 
-Step 6 is why the `Closes #N` lines matter twice over: they auto-close the issues *and* drive the label promotion. A release that closes no issues is a clean no-op. Labelling never aborts a ship — the release is already published by then, so a failure only warns.
+Step 7 is why the `Closes #N` lines matter twice over: they auto-close the issues *and* drive the label promotion. A release that closes no issues is a clean no-op. Labelling never aborts a ship — the release is already published by then, so a failure only warns.
 
-If ship.sh fails at any step, it prints what to do. Fix on dev, commit, re-run — it picks up where it left off.
+If ship.sh fails at any step, it prints what to do. Fix on dev, commit, re-run — it picks up where it left off. That includes the post-merge window: if a run dies after merging but before the tag is pushed, main is bumped but nothing is published, and a plain re-run detects the untagged release and finishes it.
 
 ## Red Flags
 
--- a/.claude/skills/ship-release/SKILL.md
+++ b/.claude/skills/ship-release/SKILL.md
@@ -154,7 +154,7 @@ ship.sh will:
 
 Step 7 is why the `Closes #N` lines matter twice over: they auto-close the issues *and* drive the label promotion. A release that closes no issues is a clean no-op. Labelling never aborts a ship — the release is already published by then, so a failure only warns.
 
-If ship.sh fails at any step, it prints what to do. Fix on dev, commit, re-run — it picks up where it left off. That includes the post-merge window: if a run dies after merging but before the tag is pushed, main is bumped but nothing is published, and a plain re-run detects the untagged release and finishes it.
+If ship.sh fails at any step, it prints what to do. Fix on dev, commit, re-run — it picks up where it left off. That includes the post-merge window: if a run dies after merging but before the tag is pushed, main is bumped but nothing is published. Preflight checks for that state on every run — main's version having no matching tag, with a merged dev → main PR behind it — and finishes the release before considering new work.
 
 ## Red Flags
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -60,6 +60,16 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           VERSION=$(node -p "require('./package.json').version")
+
+          # The tag is now the input that armed this run, so it is a second
+          # source of truth alongside package.json. ship.sh keeps them in step;
+          # a hand-pushed tag might not, and a mismatch would create a stray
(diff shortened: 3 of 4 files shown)


garretpremo (commented):

All seven acceptance criteria from #128 are met, and I was able to empirically resolve the one thing the PR body flags as "not verifiable pre-merge" — the annotated-tag `github.sha` semantics hold, so the curated release body and ship.sh's run discovery both work on a tag push. No blocking issues; the observations below are hardening, not defects in what was asked for.

<details>
<summary>Verification I ran (de-risks the "not verifiable pre-merge" note)</summary>

The whole design rests on `github.sha` / the run's `head_sha` being the **commit** sha on an annotated-tag push (ship.sh creates annotated tags via `git tag -a`). If it were the tag-object sha instead, `gh api commits/${{ github.sha }}/pulls` would return empty (release notes silently fall back to a commit dump) *and* `gh run list --commit "$MERGE_SHA"` would never match (ship.sh would hang 300s then hard-fail on every release). Checked against two repos that push annotated tags:

| repo | tag | ref object type | dereferenced commit | run `head_sha` |
|---|---|---|---|---|
| `vitejs/vite` | `v8.2.0-beta.0` | `tag` | `34af7b76…` | `34af7b76…` |
| `sindresorhus/got` | `v14.6.4` | `tag` | `edc83484…` | `edc83484…` |

Both match the commit, not the tag object. The assumption holds.

Two other things that could have silently blocked a tag-triggered publish, both clear:

- `repos/normalled/apijack/environments/release` → `protection_rules: []`, `deployment_branch_policy: null`. No deployment-branch/tag restriction that would gate a run from `refs/tags/*`.
- No tag rulesets on the repo (`/rulesets` has only branch-target rules; `/tags/protection` 404s), so `git push origin vX.Y.Z` from ship.sh is unrestricted and — being a user push, not `GITHUB_TOKEN` — does trigger workflows.

Also reproduced the `set -e` claim from the test plan with a stub: `git rev-parse … && return 0` followed by `[ -n "$(…)" ] && return 0` falls through to the end of the function without exiting, under `set -euo pipefail`. And confirmed `gh run list -c/--commit` is a real flag on the installed gh.

npm trusted publishing binds to repo + workflow filename + environment, all unchanged, so OIDC still matches on a tag-triggered run.

</details>

<details>

- **The `resolve_merge_sha` failure path advertises a resume that won't fire.** It prints "Re-run ./scripts/ship.sh — it will detect the untagged release and finish it", but `resume_untagged_release` is only reached when preflight computes `COMMITS = 0` (`scripts/ship.sh:181`), and that comparison uses the *local* `origin/main` ref. Nothing fetches `origin/main` between the version bump and `tag_and_publish` — `git push -u origin dev` updates only `origin/dev`, and `gh pr merge` is server-side. So if a run dies inside `resolve_merge_sha` (the one step before `tag_and_publish`'s own `git fetch origin main`), the local `origin/main` is still pre-merge, `COMMITS > 0`, and the re-run instead bumps to the next patch and ships that — leaving the merged version permanently unpublished. Every *other* failure point in `tag_and_publish` is after the fetch and resumes correctly. One-line fix: `git fetch origin main --tags --quiet` immediately before the `COMMITS=` line in preflight.
- **The resume only fires while `dev` has nothing ahead of `main`.** If a half-finished ship is followed by any commit to `dev`, `COMMITS != 0` and the untagged release is stranded for good — the next ship just bumps past it. Probably acceptable given how narrow the window is, but the SKILL.md wording ("a plain re-run detects the untagged release and finishes it") reads as unconditional.
- **`Create GitHub release` still derives the tag name from `package.json` rather than the ref that triggered the run.** With the trigger now being the tag itself, `${{ github.ref_name }}` is the authoritative name and `VERSION=$(node -p …)` is a second, independent source. They agree when ship.sh does the tagging; if they ever disagree (a hand-pushed `v1.16.0` at a commit whose package.json says `1.15.0`), `gh release create "v$VERSION"` would create a *second* tag at that commit and `npm publish` would collide on the already-published version. Using `github.ref_name` for the release tag would make the step self-consistent with what actually armed it.
- Removing `main` from `on.push.branches` leaves no orphaned consumers — `release-notify-issues.yml` keys off `release.published`, and `ci.yml` / `e2e.yml` carry their own `push: branches: [main, dev]`. Confirmed, just noting it was checked.
- `docs/automation-audit.md:130` still lists the publish trigger as "push to main with `chore(release):`", and line 202 attributes the tagging to `publish.yml`. The file is an explicitly dated snapshot (2026-04-27), so leaving it is defensible — flagging only because that table is the kind of thing that gets read later as current.
- Tests / E2E didn't run on this PR: the `paths` filters in `ci.yml` and `e2e.yml` don't cover `.github/workflows/publish.yml`, `scripts/**`, or `.claude/**`. That's the filters working as designed for a change with no `src/**` surface, not a red check — noting it so the empty checks list isn't mistaken for a gap.

</details>

<details>
<summary>Nitpicks</summary>

- `resume_untagged_release` sets `NEW_VERSION="$main_version"` but never reads it — `tag_and_publish` is called with `"$main_version"` directly and the trailing `ok` line interpolates `$main_version` too.
- `tag_and_publish` declares `local sha="$1" version="$2"` and then builds the tag from the positional again: `local tag="v$2"`. `"v$version"` on the next line would match the style it just established.
- `gh pr list --base main --head dev --state merged --limit 1` leans on gh's default created-desc ordering to pick "the release PR". Deterministic in practice, but the ordering is implicit.

</details>

PR #137, opened 2026-08-03
fix: skip alias validation before codegen has run (#136)

garretpremo:
Closes #136

## Summary

Before `generate` has run, a consumer's `aliases.json` points at generated commands that aren't registered yet — so `rewriteArgv` reported *every* alias as broken, one stderr line each, on *every* invocation:

```
$ rrc --version
apijack: aliases: alias "matter list" → "matter matter-list" does not resolve to a known command
apijack: aliases: alias "matter create" → "admin matters create-matter" does not resolve to a known command
... 8 more ...
1.15.0
```

That's the state every fresh install starts in (`.apijack/generated/` is gitignored by design, so standalone-binary distributions can never ship it), and it lands on the worst possible surface: `setup`, a new user's very first command. Not a correctness bug — the aliases were right the whole time, and go silent once codegen runs.

This takes the issue's primary suggestion rather than the coarse fallback. `cli-builder.ts` already knows whether the generated command surface exists, so "codegen hasn't run" is an exact signal, not a name-based heuristic: when generated commands aren't registered, unresolved expansions are skipped silently — there is nothing meaningful to validate against. When they *are* registered, a genuinely broken alias still errors exactly as before.

### `src/aliases.ts`

`rewriteArgv` takes a fourth parameter, defaulting to validation-on so the change is purely additive for every existing caller:

```ts
opts?: { generatedCommandsPresent?: boolean }
```

When `false`, an expansion missing from `realPaths` is dropped from the valid map **silently** instead of pushing an error. Everything else is untouched: the shadow-warning branch, longest-prefix matching, tail passthrough, and the return shape are all unchanged, and `continue` runs on both paths — no alias that previously applied stops applying, and none previously dropped now applies.

### `src/cli-builder.ts`

The single call site opts in, gated on both halves of the registration condition:

```ts
{ generatedCommandsPresent: commandsModule !== null && ApiClientClass !== null }
```

Both are required because registration itself requires both. `commandsModule` is assigned as soon as `generated/commands` imports, but the `generated/client` import follows it inside the same `try` — so a partial codegen (valid `commands.ts`, missing `client.ts`) would otherwise leave `commandsModule` non-null while no generated commands were ever registered, resurrecting the full spam in exactly the broken state this PR is meant to quiet.

### Tests

Five unit tests on `rewriteArgv` (both states, plus the omitted-option default), and one integration test that drives a real `cli.run()` with an `aliases.json` pointing at a generated operation and no `generated/` on disk, asserting the noise never reaches stderr.

### Docs

The alias resolution rules in `CLAUDE.md` note the pre-codegen exception.

## Acceptance criteria

- [x] Pre-codegen, aliases pointing at generated commands produce zero stderr output
- [x] Pre-codegen, an alias expanding to a built-in still rewrites correctly — the fix does not disable aliasing wholesale
- [x] Post-codegen, a genuinely broken alias still emits its per-alias error
- [x] Shadow warnings (`alias "X" shadows a real command`) are unaffected in both states
- [x] The new parameter defaults to validation-on, so existing callers are unaffected
- [x] Unit coverage for the new gate in both states, plus integration coverage of the call site

## Test plan

- [x] `bun test` — 989 pass / 0 fail
- [x] Mutation-checked the integration test: hardcoding `generatedCommandsPresent: true` makes it fail with `testcli: aliases: alias "cs" → "customers list-customers" does not resolve to a known command`, confirming it actually guards the fix rather than passing vacuously.

## Reviewer notes

Two judgment calls worth a second opinion:

- **Silent suppression also silences a typo'd alias pointing at a built-in.** `{"cfg": "config swtich"}` errors post-codegen but is silent pre-codegen, so whether you learn about the typo depends on whether `generate` has run. The user still gets Commander's `error: unknown command 'cfg'` at parse time, so it's not a silent no-op — and the alternative (per-alias "does this look generated?" heuristics) is exactly what the issue rules out.
- **`resolveLeadingTokens` is deliberately untouched.** It never validated against the command tree (see its JSDoc), so it behaves identically pre- and post-codegen and needs no equivalent gate.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -265,6 +265,7 @@ Resolution rules:
 - Longest-prefix wins when multiple aliases could match.
 - An alias that shadows a real command path emits a startup warning and the real command keeps winning.
 - An expansion that doesn't resolve to a known command emits a startup error; the CLI continues without that alias applied.
+- Before `generate` has run (no generated commands registered), unresolved expansions are skipped silently instead — there is nothing meaningful to validate against yet.
 - A global `~/.<cliName>/aliases.json` is also consulted; project-local entries override global entries on conflict.
 - **Routines and MCP tool resolution use canonical names only.** Aliases are a CLI ergonomics layer; routine YAML and MCP tool names are not affected.
 
--- a/src/aliases.ts
+++ b/src/aliases.ts
@@ -157,7 +157,10 @@ export interface RewriteResult {
  * - Aliases that match a real command path are skipped (warning emitted, real
  *   command keeps winning).
  * - Aliases whose expansion does not resolve to a real command path are skipped
- *   (error emitted, CLI continues without the alias).
+ *   (error emitted, CLI continues without the alias) -- unless
+ *   `opts.generatedCommandsPresent` is `false`, in which case they are skipped
+ *   silently, since there is nothing meaningful to validate against before
+ *   `generate` has run.
  * - Longest-prefix wins: multi-token aliases are matched before shorter ones.
  * - Trailing args (positional and flags) are appended verbatim.
  */
@@ -165,7 +168,9 @@ export function rewriteArgv(
     args: string[],
     aliases: AliasMap,
     realPaths: Set<string>,
+    opts?: { generatedCommandsPresent?: boolean },
 ): RewriteResult {
+    const generatedCommandsPresent = opts?.generatedCommandsPresent ?? true;
     const warnings: string[] = [];
     const errors: string[] = [];
     const valid = new Map<string, string[]>();
@@ -179,9 +184,12 @@ export function rewriteArgv(
         }
 
         if (!realPaths.has(expansion)) {
-            errors.push(
-                `alias "${alias}" → "${expansion}" does not resolve to a known command`,
-            );
+            if (generatedCommandsPresent) {
+                errors.push(
+                    `alias "${alias}" → "${expansion}" does not resolve to a known command`,
+                );
+            }
+
             continue;
         }
 
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -949,7 +949,10 @@ export function createCli(options: CliOptions): Cli {
 
             // 13. Apply project-local aliases (.apijack/aliases.json) by rewriting argv
             // before Commander parses. Real command paths always win over aliases;
-            // expansions that don't resolve to a known command are skipped with an error.
+            // expansions that don't resolve to a known command are skipped with an error
+            // — unless codegen hasn't run yet (or ran incompletely), in which case
+            // there's nothing meaningful to validate against and unresolved expansions
+            // are skipped silently.
             // The map was already loaded at the top of run() for early best-effort
             // resolution of pre-build argv reads — we reuse it here for the validated
             // rewrite now that the full Commander tree exists.
@@ -959,7 +962,7 @@ export function createCli(options: CliOptions): Cli {
                     process.argv.slice(2),
                     aliasMap,
                     realPaths,
-                    { generatedCommandsPresent: commandsModule !== null },
+                    { generatedCommandsPresent: commandsModule !== null && ApiClientClass !== null },
                 );
 
                 for (const w of warnings) {
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -959,6 +959,7 @@ export function createCli(options: CliOptions): Cli {
                     process.argv.slice(2),
                     aliasMap,
                     realPaths,
+                    { generatedCommandsPresent: commandsModule !== null },
                 );
 
                 for (const w of warnings) {
--- a/src/cli-builder.ts
+++ b/src/cli-builder.ts
@@ -962,7 +962,10 @@ export function createCli(options: CliOptions): Cli {
                     process.argv.slice(2),
                     aliasMap,
                     realPaths,
-                    { generatedCommandsPresent: commandsModule !== null && ApiClientClass !== null },
+                    // Mirrors the registration condition above — an existing client
+                    // module that doesn't export ApiClient leaves ApiClientClass
+                    // `undefined`, not `null`.
+                    { generatedCommandsPresent: !!commandsModule && !!ApiClientClass },
                 );
 
                 for (const w of warnings) {
--- a/tests/aliases.test.ts
+++ b/tests/aliases.test.ts
@@ -250,6 +250,60 @@ describe('rewriteArgv()', () => {
         const result = rewriteArgv(['config', 'cs'], aliases, realPaths);
         expect(result.rewrittenArgs).toEqual(['config', 'cs']);
     });
+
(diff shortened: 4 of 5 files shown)


garretpremo (commented):

Looks good — the fix takes the issue's primary suggestion (exact "codegen hasn't run" signal rather than a name-based heuristic), keeps the new parameter additive with validation on by default, and every acceptance criterion is covered by a test. CI is green across all 10 checks.

Verified locally on `01ff015`: `bun test tests/aliases.test.ts tests/cli-builder.test.ts` → 70 pass / 0 fail, and the integration test's mutation check reproduces as claimed — hardcoding `generatedCommandsPresent: true` fails with `testcli: aliases: alias "cs" → "customers list-customers" does not resolve to a known command`, so the test genuinely guards the fix.

<details>

- `src/cli-builder.ts:965` uses `commandsModule !== null && ApiClientClass !== null`, but the registration condition at `src/cli-builder.ts:646` is a truthiness check (`if (commandsModule && ApiClientClass)`). These diverge in one degenerate state: if `generated/client.ts` imports successfully but doesn't export `ApiClient`, `ApiClientClass` is `undefined` — `undefined !== null` is `true`, so the guard would report generated commands as present while registration was skipped, resurrecting the full spam. That's the exact partial-codegen class of state the PR body cites as the reason for gating on both halves. A missing `client.ts` (the case described) is handled correctly, since the throw leaves it `null`. Mirroring line 646's truthiness (`!!commandsModule && !!ApiClientClass`) would make the two conditions identical by construction rather than by coincidence.
- For a consumer that never runs `generate` at all (pure custom commands / dispatchers via `.apijack/`), alias validation is now permanently off — a typo'd alias pointing at a built-in is silent forever, not just until codegen. This is exactly what the issue asked for ("skip alias validation entirely" when no generated commands exist) and the PR body flags the tradeoff, so no change requested; noting it as the behavioral surface that got wider than "first run only."
- `CLAUDE.md` is the only place the alias resolution rules are documented (no README/skills copy to keep in sync), so the doc update is complete.

</details>

<details>
<summary>Nitpicks</summary>

- The `builtinPaths` const in `tests/aliases.test.ts` is declared mid-`describe` between tests rather than at the top of the block. It reads fine with the comment explaining why it differs from the shared `realPaths`, but hoisting it above the first pre-codegen test would match the file's existing layout.

</details>

garretpremo (commented):

Re-reviewed on `b49115e`. The guard divergence flagged last cycle is genuinely fixed — `src/cli-builder.ts:968` now reads `!!commandsModule && !!ApiClientClass`, character-for-character the truthiness of the registration condition at `src/cli-builder.ts:646`, so the partial-codegen state (a `client.ts` that imports but doesn't export `ApiClient`) can no longer report generated commands as present while registration is skipped. All six acceptance criteria are met and each is backed by a test. CI is green on the head commit — all 9 checks.

Verified locally on `b49115e`:

- `bun test tests/aliases.test.ts tests/cli-builder.test.ts` → 70 pass / 0 fail; `bun run lint` → 0 errors, 118 warnings (matches the stated baseline).
- Mutation check on the integration test reproduces as claimed: hardcoding `generatedCommandsPresent: true` fails with `testcli: aliases: alias "cs" → "customers list-customers" does not resolve to a known command`, so the test guards the fix rather than passing vacuously.
- Full suite was 988 pass / 1 fail, but the single failure (`getPluginPaths() > sourceDir points to project root`) is an artifact of my running from a git worktree with a symlinked `node_modules` — that test asserts the resolved source dir equals the project root. It passes in a normal checkout and is green on CI for both ubuntu and windows. Unrelated to this PR.

<details>

- The `b49115e` fix itself has no regression coverage. Mutating the guard back to the previous `commandsModule !== null && ApiClientClass !== null` leaves the entire suite green (989 pass / 0 fail) — nothing exercises the degenerate state where `generated/client.ts` imports cleanly but omits the `ApiClient` export. The correctness argument now rests on the two conditions being textually identical rather than on a test, which is a reasonable place to land for a state that only arises from broken codegen output; noting it so a future refactor of either line doesn't silently re-widen the gap without a failing test to catch it.
- `rewriteArgv` is not re-exported from `src/index.ts`, so the fourth parameter is an internal signature change with exactly one call site. The additive default is still the right call for the unit tests' sake, but there's no published-API constraint riding on it — worth knowing if the option ever wants to become required.
- Post-codegen shadow-warning behavior (criterion 4's other half) is covered by the pre-existing tests at `tests/aliases.test.ts:207` and `:243`, which run with the option omitted and therefore on the validation-on path. Coverage is complete across both states even though only the pre-codegen case got a new test.

</details>

<details>
<summary>Nitpicks</summary>

- `builtinPaths` is now hoisted as asked, but the first of the three pre-codegen tests (`alias pointing at a generated path is skipped silently`) still builds an inline `new Set(['generate', 'config'])` while its two siblings use the shared const. The inline set is a strict subset, so behavior is identical — using `builtinPaths` there too would make the three read as one group.

</details>

garretpremo (commented):

Approved.

- The `b49115e` fix itself has no regression coverage. Mutating the guard back to the previous `commandsModule !== null && ApiClientClass !== null` leaves the entire suite green (989 pass / 0 fail) — nothing exercises the degenerate state where `generated/client.ts` imports cleanly but omits the `ApiClient` export. The correctness argument now rests on the two conditions being textually identical rather than on a test, which is a reasonable place to land for a state that only arises from broken codegen output; noting it so a future refactor of either line doesn't silently re-widen the gap without a failing test to catch it. (#138)
- `rewriteArgv` is not re-exported from `src/index.ts`, so the fourth parameter is an internal signature change with exactly one call site. The additive default is still the right call for the unit tests' sake, but there's no published-API constraint riding on it — worth knowing if the option ever wants to become required.
- Post-codegen shadow-warning behavior (criterion 4's other half) is covered by the pre-existing tests at `tests/aliases.test.ts:207` and `:243`, which run with the option omitted and therefore on the validation-on path. Coverage is complete across both states even though only the pre-codegen case got a new test.

garretpremo (commented):

Approved.

- The `b49115e` fix itself has no regression coverage. Mutating the guard back to the previous `commandsModule !== null && ApiClientClass !== null` leaves the entire suite green (989 pass / 0 fail) — nothing exercises the degenerate state where `generated/client.ts` imports cleanly but omits the `ApiClient` export. The correctness argument now rests on the two conditions being textually identical rather than on a test, which is a reasonable place to land for a state that only arises from broken codegen output; noting it so a future refactor of either line doesn't silently re-widen the gap without a failing test to catch it. (#138)
- `rewriteArgv` is not re-exported from `src/index.ts`, so the fourth parameter is an internal signature change with exactly one call site. The additive default is still the right call for the unit tests' sake, but there's no published-API constraint riding on it — worth knowing if the option ever wants to become required.
- Post-codegen shadow-warning behavior (criterion 4's other half) is covered by the pre-existing tests at `tests/aliases.test.ts:207` and `:243`, which run with the option omitted and therefore on the validation-on path. Coverage is complete across both states even though only the pre-codegen case got a new test.

PR #139, opened 2026-08-03
chore: fire release notifications from publish.yml (#133)

garretpremo:
Closes #133

## Summary

`.github/workflows/release-notify-issues.yml` listened on `release: published` and was meant to comment "Shipped in vX 🚀" on every closed issue in a release's commit range. It has **never fired** — not once. The release is created inside `publish.yml` with `secrets.GITHUB_TOKEN`, and events raised by `GITHUB_TOKEN` do not trigger further workflow runs.

Worth correcting the record: this was previously attributed to the workflow file "not having reached `main`". It did reach `main` (commit `4ba48be`), and v1.13.0, v1.14.0, v1.14.1, v1.14.2 and v1.15.0 all shipped with it present — still zero runs. The token is the cause, not file placement.

This PR ports the logic to `scripts/notify-shipped-issues.sh`, calls it inline from `publish.yml` after `npm publish`, and deletes the dead workflow.

### New: `scripts/notify-shipped-issues.sh`

```
scripts/notify-shipped-issues.sh <tag> <release-url> [--dry-run]
```

Behavior is ported from the deleted workflow — resolve the previous stable tag, scan `PREV_TAG..TAG` commit subjects and bodies for `#N`, skip pull requests and non-closed issues, comment `Shipped in [<tag>](<url>) 🚀` — with three additions:

- **Idempotence guard.** Existing comments are checked for a `Shipped in [<tag>](` marker before posting, so a manual or repeated invocation can't double-comment. The marker is tag-anchored, so `v1.1.0` does not suppress `v1.11.0`.
- **`--dry-run`** prints what it would post and posts nothing.
- **Argument validation** rejects an empty tag or empty release URL and errors on unknown flags, rather than absorbing them as positionals.

A per-issue comment failure logs to stderr and continues; the script's only non-zero exits are argument validation and a genuine `git` failure.

### Modified: `.github/workflows/publish.yml`

`issues: write` added to the stable `publish` job only, plus a final step:

```yaml
      - name: Notify shipped issues
        continue-on-error: true
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          REPO: ${{ github.repository }}
        run: |
          RELEASE_URL=$(gh release view "${{ github.ref_name }}" --json url --jq .url)
          scripts/notify-shipped-issues.sh "${{ github.ref_name }}" "$RELEASE_URL"
```

It runs last, after `npm publish`, so "shipped" means npm actually published. `continue-on-error: true` means a notification failure can never fail a release. `publish-next` deliberately gets no notify step — pre-releases don't notify. The old workflow's `if: !github.event.release.prerelease` guard is subsumed by the job's existing gate (`startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-')`), which is strictly narrower.

### Not redundant with `promote-shipped-issues.sh`

Option 4 in the issue was to delete this if `scripts/promote-shipped-issues.sh` already covered it. It doesn't: that script applies the `deployed` label based on `Closes #N` refs in the release PR body and posts no comment. This one posts a comment based on the whole commit range. Complementary, and the script header says so.

## Acceptance criteria

- [x] Determined the notification is still wanted and is not redundant with `promote-shipped-issues.sh`
- [x] Notify logic actually fires on a release — runs inline in `publish.yml` rather than via an event `GITHUB_TOKEN` cannot raise
- [x] Dead `release-notify-issues.yml` removed
- [x] Runs for stable releases only; `publish-next` unaffected
- [x] A notification failure cannot fail an otherwise-successful release
- [x] A re-run does not double-comment
- [x] A broken revision range fails loudly instead of reading as "no issues to notify" — the silent-failure mode that hid the original bug for five releases

## Test plan

- [x] `bun test` — 998 pass / 0 fail
- [x] `bun run lint` — 0 errors (118 pre-existing warnings, none in the new files)
- [x] `tests/notify-shipped-issues.test.ts` — 15 tests against the real script in a temp git fixture repo with a PATH-stubbed `gh`: no-previous-tag, no-references, PR skip, open-issue skip, closed-issue comment body, duplicate-comment skip, cross-tag marker discrimination (`v1.1.0` vs `v1.11.0`), comment failure on one issue not aborting the rest, `--dry-run`, empty/missing/unknown arguments, and broken revision range
- [x] Test honesty verified by mutation — reviewer independently reverted each fix and confirmed the covering test goes red, including the broken-range test and the idempotence guard
- [x] `scripts/notify-shipped-issues.sh v1.15.0 <url> --dry-run` against this repo — lists plausible issue numbers, posts nothing
- [x] `publish.yml` parses as YAML; notify step confirmed inside `publish`, not `publish-next`

The issue's third criterion — "verify with a real release that the comment appears" — can only close on the next actual release. Everything short of that is covered above.


garretpremo (commented):

This does what #133 asked for: the notify logic now runs as a step inside `publish.yml` rather than behind a `release: published` event that `GITHUB_TOKEN` can never raise, the dead workflow is gone, and the new script is covered by 15 tests against a real git fixture repo. All seven acceptance criteria in the PR body check out, and the diff stays inside the issue's scope.

Verified independently:

- `publish`'s checkout is `fetch-depth: 0`, so `git tag` / `git log "$PREV_TAG..$TAG"` see the full history the script needs — this was the most likely way for the port to reproduce the original silent no-op, and it doesn't.
- Step ordering is right: `Create GitHub release` → `npm publish` → `Notify shipped issues`, so the release exists when `gh release view` runs and "shipped" means npm actually published.
- `gh release view <tag> --json url` returns the browser URL (`https://github.com/normalled/apijack/releases/tag/v1.15.0`), not `apiUrl` — the comment link resolves for a human.
- Stable-only holds via the job's existing `startsWith(github.ref, 'refs/tags/v') && !contains(github.ref, '-')` gate, which is strictly narrower than the deleted workflow's `!prerelease` check. `publish-next` has no notify step and no `issues: write`.
- `issues: write` is scoped to the `publish` job only.
- Sourcing `gh-pin-account.sh` is a no-op in CI (`gh auth token --user` fails, `|| true` swallows it, `GH_TOKEN` from the step env survives) — the fake `gh` in the tests models this correctly.
- `scripts/notify-shipped-issues.sh v1.15.0 <url> --dry-run` against this repo: scans `v1.14.2..v1.15.0`, flags #122 to comment, skips #123/#124 as pull requests, posts nothing. No false positives in that range.
- The 15 new tests pass locally; CI is green on both `ubuntu-latest` and `windows-latest` (the suite is `describe.skipIf(win32)`, consistent with `tests/extract-closing-refs.test.ts` and #129).
- No remaining references to `release-notify-issues.yml` anywhere in the repo.

The `set -e` hardening is the part that matters most and it's real: capturing `git tag` and `git log` into variables before filtering means a broken revision range aborts instead of reading as "no issue references found" and exiting 0 — the exact shape of silent failure that hid the original bug for five releases.

<details>

- **`#N` matching is keyword-less and not fence-aware.** `grep -oE '#[0-9]+'` over `%s %b` will match a number quoted inside a fenced code block in a commit body, unlike the sibling extractors `scripts/collect-closes-refs.sh` and `scripts/extract-closing-refs.sh`, which were deliberately taught to strip fences in #129/#130. This is ported behavior and the breadth is intentional per the script header (catch mentions that never reached the curated PR body), so it isn't a regression — but note this logic has never actually executed before, so the next release is the first time a false positive can post a real comment. Blast radius is a stray "Shipped in vX 🚀" on an unrelated closed issue, which is cheap to undo. The v1.15.0 dry-run showed no noise, so this looks fine at current commit-message conventions.

- **The idempotence guard reads the oldest 100 comments, not the newest.** `GET /repos/{owner}/{repo}/issues/{n}/comments` sorts ascending by id and accepts only `since` / `per_page` / `page`, so `per_page=100` widens the window but doesn't bias it toward recent comments the way the inline rationale implies. Past 100 comments on an issue the guard fails open and double-comments. Not a concern at this repo's scale, and the fail-open direction is the right choice — just worth knowing the comment's reasoning is slightly off from what the endpoint does.

- **Issue criterion 3 stays open by construction.** "Verify with a real release that the comment appears" can only close on the next stable tag push, which the PR states plainly. Since the step is `continue-on-error: true`, a failure will surface as a failed-step annotation on a green job rather than a red run — worth actually opening the next publish run and checking that step rather than inferring success from the job's status.

</details>

<details>
<summary>Nitpicks</summary>

- Argument handling is asymmetric: an unknown flag errors, but extra positionals are silently absorbed (`notify-shipped-issues.sh v1.16.0 <url> junk` runs normally and ignores `junk`). Given the validation effort spent on the flag path, a `-gt 2` check would round it out.

- `sort -u` without `-n` orders issue numbers lexicographically (#10, #100, #11), so the log output reads oddly on a release touching many issues. Cosmetic only, and ported as-is from the deleted workflow.

</details>

garretpremo (commented):

Approved.

- **`#N` matching is keyword-less and not fence-aware.** `grep -oE '#[0-9]+'` over `%s %b` will match a number quoted inside a fenced code block in a commit body, unlike the sibling extractors `scripts/collect-closes-refs.sh` and `scripts/extract-closing-refs.sh`, which were deliberately taught to strip fences in #129/#130. This is ported behavior and the breadth is intentional per the script header (catch mentions that never reached the curated PR body), so it isn't a regression — but note this logic has never actually executed before, so the next release is the first time a false positive can post a real comment. Blast radius is a stray "Shipped in vX 🚀" on an unrelated closed issue, which is cheap to undo. The v1.15.0 dry-run showed no noise, so this looks fine at current commit-message conventions. (#140)

- **The idempotence guard reads the oldest 100 comments, not the newest.** `GET /repos/{owner}/{repo}/issues/{n}/comments` sorts ascending by id and accepts only `since` / `per_page` / `page`, so `per_page=100` widens the window but doesn't bias it toward recent comments the way the inline rationale implies. Past 100 comments on an issue the guard fails open and double-comments. Not a concern at this repo's scale, and the fail-open direction is the right choice — just worth knowing the comment's reasoning is slightly off from what the endpoint does. (#140)

- **Issue criterion 3 stays open by construction.** "Verify with a real release that the comment appears" can only close on the next stable tag push, which the PR states plainly. Since the step is `continue-on-error: true`, a failure will surface as a failed-step annotation on a green job rather than a red run — worth actually opening the next publish run and checking that step rather than inferring success from the job's status. (#140)

PR #141, opened 2026-08-03
fix: measure fence indent from the container content column (#134)

### LINKED ISSUE #100: RoutineResult: emit `steps[]` entry for `forEach` non-array path; tag iterations with index

garretpremo:
Closes #134

## Summary

`scripts/extract-closing-refs.sh` stripped fenced code blocks before matching closing keywords, but only recognized a fence opener at indent 0–3. CommonMark measures fence indent relative to the enclosing list item's **content column**, so a fence nested inside a list legitimately sits at 4+ spaces — and the extractor read straight through it:

```
$ printf 'Closes #7\n\n- outer\n  - inner\n    ```\n    Closes #100\n    ```\n' \
    | ./scripts/extract-closing-refs.sh -
7
100      <-- inside a real fence
```

That matters because the consumers mutate labels: `promote-shipped-issues.sh` *strips* `ready-for-implement` and `merged to dev` on every number it sees, and `collect-closes-refs.sh` writes matches into release PR bodies as `Closes #N`. Release PR bodies are nested-bullet-heavy with pasted YAML — exactly the shape that leaks.

The fix is one rule: **a fence opener — and a list marker — is recognized only within 3 spaces of the current container's content column, not of column 0.** Because an indented code block starts at `base + 4` by definition, that same rule makes a fence marker sitting inside indented code inert for free. No separate "am I in indented code" state exists, and none is needed.

That distinction is the whole reason #130 deferred this. Permitting a fence at *any* indent would have let a stray fence marker in an indented block open a phantom fence and swallow every real reference after it — trading a rare false positive for a broad false **negative**, which is an issue that silently never auto-closes and nothing flags.

### `scripts/extract-closing-refs.sh`

The awk pre-pass now carries block context: a stack of open list-item content columns, plus the indent and container depth each fence was opened at.

- **List markers push a content column.** `-`, `*`, `+`, and ordered `N.` / `N)` markers push `indent + markerWidth + spacesAfter`. Recognized only at `indent <= base + 3`, so a `- item` line inside an indented code block can't open a phantom container and re-enable deep fences.
- **CommonMark's 5+-space rule is honored**: a marker followed by five or more spaces caps the content column at `marker + 1`, the extra spaces being indented code inside the item. Without the cap, `base` overshoots and a fence opens where CommonMark says code — the false-negative direction.
- **Containers pop aggressively** — any non-blank line at a lower indent pops. Popping early lowers `base`, which makes a deep fence *less* likely to be recognized; the bias is deliberate.
- **A fence closes when its list item ends.** A non-blank line indented below the fence's container closes the fence and is then processed normally. Without this, a fence left open inside a list item would swallow every reference in the rest of the body.
- **Closers** match on fence char, run length `>= fence_len`, blank remainder, and `indent <= fence_indent + 3` — the lenient reading, which closes sooner.
- **A tab-led line does not trigger the early close while the fence's container sits at column 4 or below.** Indent is counted in spaces, so `\tcode` would otherwise read as column 0, dedent out of the fence, and leave the genuine closer to be re-read as a fresh opener — swallowing everything after it, silently. A tab expands to column 4, so the exemption stops there: at container column 5+ a tab-led line really is dedented out, and the reference really is prose. Both sides of that boundary are pinned.

Every listed gap in the header still errs the same direction: visible reference, never hidden one. The 4+-indent entry is gone; tabs-not-expanded and lazy-paragraph-continuation are now named.

`strip_spans`, the keyword regex, CRLF tolerance, the unterminated-fence stderr warning, and the awk-outside-the-pipeline exit handling are untouched. Indented code blocks and blockquotes are still **not** stripped — that stays the #129 decision, and the new code models indented code without stripping it.

### `tests/extract-closing-refs.test.ts`

The pinned gap test `a fence indented 4+ spaces is not recognized` moves out of `documented gaps` and into `code is not a reference`, asserting the fixed result. Sixteen cases added, each pinning a direction the state machine could fail in: single-level and nested list fences, ordered markers, a fence marker inside indented code, a list marker inside indented code, indented code *inside* a list item, a closer indented deeper than its opener, a deeper fence marker inside an open fence, a fence left open when its list ends, a fence after the list has ended, the 5+-space marker cap, thematic breaks not pushing a container, a reference carried on the marker line itself (`- Closes #7`, `1. Closes #7`, `- [ ] Closes #7`), and both sides of the tab boundary. `documented gaps` gains an explicit pin that indented code is still not stripped.

## Acceptance criteria

- [x] A fence nested inside a list item (4+ spaces of indent) has its contents ignored
- [x] An indented code block containing a fence marker does **not** open a phantom fence — references after it are still matched
- [x] The existing `documented gaps` test for this behavior asserts the fixed behavior rather than pinning the leak
- [x] No regression against the #126 and #80 bodies, or the full-repo sweep

## Test plan

- [x] `bun test` — 1000 pass / 0 fail
- [x] `bash -n scripts/extract-closing-refs.sh` — clean
- [x] Portability matrix, the script's 42 tests run against each awk via a PATH shim: **GNU awk 5.2.1** 42/0, **mawk** 42/0, **busybox awk** 42/0. `gawk --posix` is 40/42, the same two pre-existing `/dev/stderr` failures reproduced identically against `dev`
- [x] Issue repro: the nested-list body above now yields `7` alone
- [x] Tab boundary swept across container columns 0–7 against reference CommonMark — strip at 0–4, visible at 5–7, matching at every row
- [x] **Full-repo sweep** — all 135 issue and PR bodies in the repo, extracted three ways: the pre-#129 raw regex, the extractor as it stands on `dev`, and the extractor on this branch.
  - **`dev`-vs-branch: 0 differences.** No live body's result changes; the fix only closes the shape #134 describes.
  - raw-vs-branch: 4 bodies, all dropping false positives only — #126 (`10 11 12 13 14 15 16 125` → `125`) and #80 (`65 70` → nothing), the two #130 identified, plus #130 (`10 14 15 125 127 129` → `127 129`) and #134 itself (`7 100` → nothing), both filed after that sweep was taken and both carrying fenced fixtures.

### DIFF
--- a/scripts/extract-closing-refs.sh
+++ b/scripts/extract-closing-refs.sh
@@ -44,6 +44,8 @@
 #     indent 0 and a tab-indented list item as a marker at column 0
 #   - a lazy paragraph continuation dedented out of its list item pops the
 #     container early, which only lowers the column a fence must beat
+#   - a fence opener sharing a line with its list marker (`- ```) is not seen,
+#     because the opener is looked for at the line indent, not past the marker
 #
 # Network-free by design: it takes text, not a PR number. Callers do their own
 # fetching, which differs between them for good reasons (the label scripts hit
@@ -142,7 +144,10 @@ function strip_spans(s,   out, i, n, run, j, closerun, found) {
 
 # How far past its own indent the content of a list item starts — the marker width
 # plus the spaces after it. 0 when the line is not a list item. A marker at end
-# of line still opens an item whose content column is one past the marker.
+# of line still opens an item whose content column is one past the marker, and so
+# does one followed by 5+ spaces: per CommonMark those extra spaces are indented
+# code inside the item, so counting them would raise the column and let a fence
+# open where code was meant — swallowing a reference, the direction #134 avoids.
 function list_offset(p,   w, n, c) {
     w = 0
     c = substr(p, 1, 1)
@@ -161,6 +166,7 @@ function list_offset(p,   w, n, c) {
     if (substr(p, w + 1, 1) != " ") return 0
     n = 0
     while (substr(p, w + n + 1, 1) == " ") n++
+    if (n > 4) n = 1
     return w + n
 }
 
--- a/scripts/extract-closing-refs.sh
+++ b/scripts/extract-closing-refs.sh
@@ -18,21 +18,32 @@
 # unrelated release PRs from months earlier.
 #
 # So: strip code before matching.
-#   - fenced blocks (``` and ~~~, per CommonMark: 3+ delimiters, up to 3
-#     spaces of indent, closing fence at least as long as the opening one)
+#   - fenced blocks (``` and ~~~, per CommonMark: 3+ delimiters, closing fence
+#     at least as long as the opening one)
 #   - inline code spans (`x`, ``x``, ...), scanned by matching delimiter-run
 #     length so ``Closes #10`` isn't half-stripped into a live reference
+#
+# Fence indent is measured from the content column of the enclosing list item,
+# not from column 0 (#134): a fence nested in a list legitimately sits at 4+
+# spaces, and reading it as prose leaked everything inside it. The awk pre-pass
+# therefore tracks open list items and allows a fence within 3 of that column.
+# List markers are capped the same way, and that cap is what makes the rule
+# safe: base + 4 is by definition an indented code block, so a fence marker or
+# a `- item` line inside indented code opens nothing. Recognizing a fence at any
+# indent instead was rejected — a stray fence marker in indented code would open
+# a phantom fence and swallow every real reference after it, and a swallowed
+# reference is the one failure nothing flags.
 #
 # Known and deliberate gaps — all of them leave a reference VISIBLE (a possible
 # false positive), never hide a real one, which is the safer direction to err:
 #   - indented code blocks and blockquotes are not stripped (per #129: a
 #     blockquote can legitimately carry a real reference)
-#   - a fence indented 4+ spaces is not recognized, so a fence nested inside a
-#     list item leaks its contents (#134). Allowing any indent would be worse:
-#     an indented-code-block fence marker would open a phantom fence and
-#     swallow every real reference after it.
 #   - code spans are scanned per line, so a span that wraps across a newline
 #     leaks its contents
+#   - tabs are not expanded to tab stops, so a tab-indented fence reads as
+#     indent 0 and a tab-indented list item as a marker at column 0
+#   - a lazy paragraph continuation dedented out of its list item pops the
+#     container early, which only lowers the column a fence must beat
 #
 # Network-free by design: it takes text, not a PR number. Callers do their own
 # fetching, which differs between them for good reasons (the label scripts hit
@@ -129,38 +140,93 @@ function strip_spans(s,   out, i, n, run, j, closerun, found) {
     return out
 }
 
-BEGIN { in_fence = 0; fence_char = ""; fence_len = 0 }
+# How far past its own indent the content of a list item starts — the marker width
+# plus the spaces after it. 0 when the line is not a list item. A marker at end
+# of line still opens an item whose content column is one past the marker.
+function list_offset(p,   w, n, c) {
+    w = 0
+    c = substr(p, 1, 1)
+    if (c == "-" || c == "*" || c == "+") {
+        w = 1
+    } else {
+        n = 0
+        while (substr(p, n + 1, 1) ~ /^[0-9]$/) n++
+        if (n > 0) {
+            c = substr(p, n + 1, 1)
+            if (c == "." || c == ")") w = n + 1
+        }
+    }
+    if (w == 0) return 0
+    if (length(p) == w) return w + 1
+    if (substr(p, w + 1, 1) != " ") return 0
+    n = 0
+    while (substr(p, w + n + 1, 1) == " ") n++
+    return w + n
+}
+
+BEGIN {
+    in_fence = 0; fence_char = ""; fence_len = 0; fence_indent = 0; fence_base = 0
+    depth = 0; base = 0
+}
 {
-    probe = $0
-    # Up to 3 spaces of indent still opens a fence; a 4th makes it an indented
-    # code block. Trimmed with a loop rather than /^ {0,3}/ because interval
-    # expressions are not portable across awk implementations.
+    # Leading spaces, counted with a loop rather than /^ */ because the count is
+    # needed and interval expressions are not portable across awk implementations.
     indent = 0
-    while (indent < 3 && substr(probe, 1, 1) == " ") {
-        probe = substr(probe, 2)
-        indent++
-    }
-    if (match(probe, /^(```+|~~~+)/)) {
-        marker = substr(probe, 1, RLENGTH)
-        ch = substr(marker, 1, 1)
-        len = RLENGTH
-        rest = substr(probe, RLENGTH + 1)
-        if (in_fence == 0) {
-            # A backtick info string cannot itself contain a backtick, so
-            # ```` `a` ```` opens nothing — it is an inline span.
-            if (ch == "`" && index(rest, "`") > 0) {
-                print strip_spans($0)
-                next
+    while (substr($0, indent + 1, 1) == " ") indent++
+    probe = substr($0, indent + 1)
+    blank = ($0 ~ /^[[:space:]]*$/)
+
+    if (in_fence) {
+        if (blank || indent >= fence_base) {
+            if (match(probe, /^(```+|~~~+)/)) {
+                ch = substr(probe, 1, 1)
+                # The closer is measured against the opener rather than the
+                # container: the lenient reading, which closes sooner.
+                if (ch == fence_char && RLENGTH >= fence_len && indent <= fence_indent + 3 &&
+                    substr(probe, RLENGTH + 1) ~ /^[[:space:]]*$/) {
+                    in_fence = 0; fence_char = ""; fence_len = 0
+                }
             }
-            in_fence = 1; fence_char = ch; fence_len = len
             next
         }
-        if (ch == fence_char && len >= fence_len && rest ~ /^[[:space:]]*$/) {
-            in_fence = 0; fence_char = ""; fence_len = 0
-            next
+        # Dedented out of the list item holding the fence, so the fence ended
+        # with it. Fall through and process this line normally — an unclosed
+        # nested fence must not swallow the rest of the body.
+        in_fence = 0; fence_char = ""; fence_len = 0
+    }
+
+    if (!blank) {
+        # A list item may contain blank lines, so only a non-blank line closes
+        # containers. Popping eagerly lowers base, which makes a deep fence
+        # LESS likely to be recognized — the false-positive direction.
+        while (depth > 0 && indent < stack[depth]) depth--
+        base = 0
+        if (depth > 0) base = stack[depth]
+
+        # Both markers below are capped at 3 past the content column of the
+        # container. base + 4 is by definition an indented code block, so a fence
+        # marker there is inert without tracking indented code separately — and
+        # the cap on list markers is what keeps a `- item` line inside indented
+        # code from opening a container that would re-enable deep fences.
+        if (indent <= base + 3) {
+            off = list_offset(probe)
+            if (off > 0) {
+                depth++
+                stack[depth] = indent + off
+                base = stack[depth]
+            } else if (match(probe, /^(```+|~~~+)/)) {
+                ch = substr(probe, 1, 1)
+                # A backtick info string cannot itself contain a backtick, so
+                # ```` `a` ```` opens nothing — it is an inline span.
+                if (ch != "`" || index(substr(probe, RLENGTH + 1), "`") == 0) {
+                    in_fence = 1; fence_char = ch; fence_len = RLENGTH
+                    fence_indent = indent; fence_base = base
+                    next
+                }
+            }
         }
     }
-    if (in_fence == 0) print strip_spans($0)
+    print strip_spans($0)
 }
 
 # A closing fence cannot carry a trailing info string, so one stray character on
--- a/scripts/extract-closing-refs.sh
+++ b/scripts/extract-closing-refs.sh
@@ -44,18 +44,20 @@
 # the state machine is there for exactly that reason. Weigh it before touching
 # the escape.
 #
-# Known and deliberate gaps — all but the one noted above leave a reference
-# VISIBLE (a possible false positive) rather than hiding a real one, which is the
-# safer direction to err:
+# Known and deliberate gaps. Most leave a reference VISIBLE (a possible false
+# positive) rather than hiding a real one, which is the safer direction to err —
+# the two that can hide one, the early-close reopen above and the setext
+# underline below, say so where they are described:
 #   - indented code blocks and blockquotes are not stripped (per #129: a
 #     blockquote can legitimately carry a real reference)
 #   - code spans are scanned per line, so a span that wraps across a newline
 #     leaks its contents
 #   - tabs are not expanded to tab stops, so a tab-indented fence reads as
-#     indent 0 and a tab-indented list item as a marker at column 0. Tab-led
-#     lines are exempt from the fence early-close above, which covers the one
-#     shape of this that hid references; a tab after leading spaces does not
-#     get that exemption
+#     indent 0 and a tab-indented list item as a marker at column 0. A tab in
+#     column 1 is exempt from the fence early-close above while the fence sits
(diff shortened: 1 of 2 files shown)


garretpremo (commented):

All four acceptance criteria are met, and I reproduced the PR's claims independently rather than taking them on faith — the container-column state machine is correct, and it improves *both* error directions relative to `dev`, which is more than #134 asked for. No blocking issues.

## What I verified

**Acceptance criteria**

| Criterion | Result |
|---|---|
| Nested-list fence (4+ spaces) has contents ignored | ✅ issue repro yields `7` alone; `dev` yields `7 100` |
| Indented-code fence marker opens no phantom fence | ✅ pinned by test + confirmed by fuzz |
| `documented gaps` test asserts fixed behavior | ✅ moved into `code is not a reference` |
| No regression on #126 / #80 / full-repo sweep | ✅ see below |

**Full-body sweep.** Pulled all issue and PR bodies from the repo (139 — a few more exist than the 135 in the description) and ran `dev` vs this branch over each: **0 differences.** The fix only touches the shape #134 describes; no live body's result changes.

**Differential fuzz against a reference CommonMark parser.** This is the check I most wanted, since the header's own argument is that a *hidden* reference is the failure nothing flags. I generated 4000 randomized bodies (nested lists, fences, indented code, blockquotes, tabs, thematic breaks, ordered/unordered markers, 5+-space markers), parsed each with `commonmark`, and compared the set of references CommonMark places in prose against the script's output:

- **False negatives (hidden references): 0.** Not one body across 4000.
- False positives: **60 → 33** vs `dev` — the safe direction, nearly halved.

On a second corpus deliberately seeded with the hostile shapes (setext headings, lone markers, marker-line fences, HTML blocks), the branch is still strictly better: false-negative bodies **817 → 487**, false positives **1956 → 1422**. The dominant false negative on `dev` is `` - ``` `` swallowing the rest of the body; this branch turns that into a *visible* gap instead. Notably the tab-reopen case in the description is real — `dev` returns `7` where CommonMark says `7 8`, and this branch returns `7 8`.

**Portability matrix.** Confirmed via a `PATH` shim: **gawk 5.2.1** 42/0, **mawk** 42/0, **busybox awk** 42/0. `gawk --posix` is 40/42, and I confirmed the two `/dev/stderr` failures are pre-existing — `old.sh` and `new.sh` produce byte-identical output on the unterminated-fence input under `--posix`.

**Other.** Full suite 1000 tests / 999 pass (the one failure is `paths.test.ts` asserting the project dir contains `apijack`, an artifact of my running from a worktree — unrelated to this change). `bun run lint` 0 errors / 118 pre-existing warnings. `bash -n` clean. CI green on all five checks. CRLF bodies still handled. Diff is confined to the two relevant files; no public API or generated-output change.

<details>

- **The two gaps that can *hide* a reference aren't pinned by tests.** The `documented gaps` preamble in the test file says these "are pinned so the behavior can't drift silently," and the header comment is careful to call out the setext underline and the early-close reopen as the two that can hide rather than expose. But neither has a test. The setext one is concretely reachable:

  ```
  Title
  -

      ```
      Closes #1
      ```
  ```

  This branch returns nothing; `dev` returns `1`. CommonMark renders that block as an *indented* code block (`<h2>Title</h2>` followed by `<pre><code>``` …`), and indented code is deliberately not stripped — so `1` should stay visible. The shape is narrow enough not to block: it needs a setext H2 underlined with a *bare lone* `-` specifically (`--` and `---` both take other paths, and `---` is caught by the thematic-break check), followed by 4-space indented content that starts with a fence marker. Zero real bodies hit it. Given the file's convention, though, a pin in `documented gaps` would keep it honest — this is the one direction the header argues nothing else will catch.

- **`- ---` is read as a thematic break, not a list item.** `list_offset` strips whitespace and sees `----` (4 identical chars, whole line), so it returns 0. CommonMark reads `- ---` as a list item *containing* a thematic break. The effect is a lower `base`, which makes a deep fence *less* likely to be recognized — the false-positive direction, so it's safe. Just slightly wider than the comment's "3+ of the same character and nothing else" describes.

</details>

<details>
<summary>Nitpicks</summary>

- The ordered-marker scan accepts an unbounded run of digits; CommonMark caps list markers at 9. `1234567890. x` would push a container CommonMark wouldn't, which nudges `base` up — technically the unsafe direction, though the shape is vanishingly unlikely in a PR body.
- In `list_offset`, the thematic-break branch tests `c == "_"`, but `_` is never a list marker, so that arm can't change the result. Harmless, and arguably clearer for mirroring CommonMark's rule set.

</details>

garretpremo (commented):

Re-review after `b058b14`. The two pins that were missing are now present and both assert honestly, the ordered-marker cap landed with a test on each side of the boundary, and I re-ran the full verification from scratch rather than carrying the previous pass forward. All four acceptance criteria hold. No blocking issues.

## What I verified

**Acceptance criteria**

| Criterion | Result |
|---|---|
| Nested-list fence (4+ spaces) has contents ignored | ✅ issue repro yields `7`; `dev` yields `7 100` |
| Indented-code fence marker opens no phantom fence | ✅ pinned by test, and confirmed by fuzz below |
| `documented gaps` test asserts fixed behavior | ✅ moved into `code is not a reference` |
| No regression on #126 / #80 / full-repo sweep | ✅ 0 diffs on 139 bodies; #126 `→ 125`, #80 `→ ∅`, both dropping false positives only |

**Full-body sweep.** All 139 issue and PR bodies in the repo, `dev` vs this branch: **0 differences.**

**Differential fuzz against a reference CommonMark parser.** 4000 randomized bodies (nested lists, fences, indented code, tabs, thematic breaks, ordered and 5+-space markers), comparing the references CommonMark places in prose against the script's output:

- False-negative bodies (a *hidden* reference — the failure the header argues nothing flags): **36 → 5**.
- False-positive bodies: **2527 → 2349**.

Strictly better in both directions. I checked the 5 remaining false negatives: they are unbalanced-fence shapes and every one fires `warning: unterminated code fence`, so they stay audible — which is exactly the property the header claims for this class.

**The new commit's specific claims.** Checked each against `commonmark`:

- `123456789.` (9 digits) opens a container and strips the fence under it; `1234567890.` (10) does not, leaving `100` visible as indented code. Both match the spec cap.
- The setext pin is real and correctly labeled: CommonMark renders `Title` / `-` / 4-space block as `<h2>` + `<pre><code>`, so `Closes #1` is *indented* code, which #129 says stays visible. The branch hides it. Asserting `[]` with empty stderr is the honest pin.
- The early-close pin's comment is accurate — CommonMark does put `Closes #7` inside a code block for that body, so extractor and spec agree there.

**Portability.** Via a `PATH` shim: **gawk 5.2.1** 45/0, **mawk** 45/0, **busybox awk** 45/0. `gawk --posix` is 42/45; I confirmed all three failures are the pre-existing `/dev/stderr` write, not new — running `dev`'s own test file under `--posix` fails the same two stderr cases.

**Other.** Full suite **1003 pass / 0 fail**. `bun run lint` 0 errors / 118 pre-existing warnings. `bash -n` clean. CI green on all five checks including `unit-tests (windows-latest)`, which was pending at the start of this review. Diff confined to the two relevant files; no public API or generated-output change.

<details>

- **The `- ---` comment added in `b058b14` states the spec backwards, and my earlier review is why.** It now reads:

  ```
  # That also catches `- ---`, which CommonMark reads as a list item CONTAINING
  # a break; returning 0 there only lowers the column, so it errs the safe way.
  ```

  I checked `- ---` against the reference parser and it renders as a bare `<hr />`, not a list item — thematic break takes precedence over list item when a line is ambiguous, and `- ---` is four `-` separated by a space, a valid break. So `list_offset` returning 0 isn't erring safe, it's simply correct, and matches CommonMark exactly. My previous review asserted the opposite and this comment was written to it; correcting the record so the wrong claim doesn't outlive it. Code unchanged either way.

- **The `documented gaps` label on the early-close pin is stronger than the case warrants.** It's headed `HIDES a reference`, but as the test's own comment notes, CommonMark also reads that `Closes #7` as code — so in that shape the extractor isn't diverging from the spec at all. The header's mechanism-level claim is still right (an early close *can* hide), and the shape it does hide in is genuinely hard to construct, which may be why this stand-in was chosen. Worth a word in the test so a later reader doesn't take the pin as evidence of a divergence it doesn't demonstrate.

- **`gawk --posix` now fails 3 tests rather than 2.** No regression — the added case exercises the same `/dev/stderr` write, and I verified `dev` fails identically on the pre-existing two. But that single stderr write is the only thing between this script and POSIX-clean, and the count of tests riding on it is growing.

</details>

<details>
<summary>Nitpicks</summary>

- My earlier nit about the `c == "_"` arm was imprecise: `_` is a legitimate *thematic break* character, and that arm sits in the thematic-break pre-check, not the marker scan. It still can't change the return value (`___` falls through to `w = 0` and returns 0 anyway), so leaving it for symmetry with CommonMark's rule set is the right call.

</details>

garretpremo (commented):

Re-review after `4d6cbd0`, which is comment-only — no logic change. I checked both corrected comments against a reference CommonMark parser rather than accepting them, and re-ran the full verification from scratch. All four acceptance criteria hold. No blocking issues.

## What I verified

**Acceptance criteria**

| Criterion | Result |
|---|---|
| Nested-list fence (4+ spaces) has contents ignored | ✅ issue repro yields `7`; `dev` yields `7 100` |
| Indented-code fence marker opens no phantom fence | ✅ `42` still visible after an indented block starting with a fence marker; matches `dev` |
| `documented gaps` test asserts fixed behavior | ✅ now at line 107, inside `code is not a reference` |
| No regression on #126 / #80 / full-repo sweep | ✅ 139 bodies, **0 diffs** vs `dev`; #126 `→ 125`, #80 `→ ∅`, raw regex on #126 still `10 11 12 13 14 15 16 125` |

**The new commit's two claims.** Both check out against `commonmark`:

- `- ---` renders as a bare `<hr />`, not a list item containing a break — thematic break outranks list item on an ambiguous line, confirmed against the spec's own `* Foo / * * * / * Bar` example. So the rewritten comment is right that `list_offset` returning 0 *matches* CommonMark rather than merely erring safe, and the previous review's claim it corrects was the wrong one.
- The early-close pin's relabel is accurate. For that body CommonMark does place `Closes #7` in a code block, so the extractor isn't diverging there — "pins the MECHANISM, not a divergence" is the honest description, and it's better than what it replaced.

**Differential fuzz against CommonMark.** 3000 randomized bodies (nested lists, fences of all three delimiters, indented code, tabs, thematic breaks, ordered and empty markers, blockquotes), comparing the references CommonMark places outside code spans and code blocks against the script's output:

- False-negative bodies (a *hidden* reference): **dev 0 → branch 0.**
- False-positive bodies: **1812 → 1780**, the safe direction. (The absolute count is inflated by my generator emitting blockquote- and indented-code-borne refs, which #129 deliberately leaves visible.)

**Portability.** Via a `PATH` shim: **gawk 5.2.1** 45/0, **mawk** 45/0, **busybox awk** 45/0.

**Other.** Full suite 1003 tests, 1002 pass — the one failure is `paths.test.ts` asserting the project dir contains `apijack`, an artifact of my running from a worktree, and CI is green on both runners regardless. `bun run lint` 0 errors / 118 pre-existing warnings. `bash -n` clean. CI green on all three checks. Diff confined to the two relevant files; no public API or generated-output change.

<details>

- **The setext gap is the one shape where this branch is strictly worse than `dev`, and that's worth stating plainly even though it shouldn't block.** For a `Title` line over a bare lone `-`, followed by a blank line and a 4-space-indented block whose first line is a fence marker: `dev` returns the reference, this branch returns nothing, silently. CommonMark renders that as an `h2` followed by an *indented* code block, and #129 says indented code stays visible — so this is a real divergence in the direction the header itself argues nothing flags, newly introduced here rather than pre-existing. What makes it non-blocking: the shape is very narrow (a bare lone `-` underline specifically, then 4-space content beginning with a fence marker), the full-repo sweep shows zero real bodies reaching it, the fuzz never generated one, and it is pinned in `documented gaps` with an accurate "HIDES a reference" label plus an explicit `expect(stderr).toBe('')` recording the silence. That's the same way #129 and #130 handled this class. Fixing it needs previous-line state the pre-pass doesn't carry, which is a fair thing to defer — but it deserves a tracking issue rather than only a code comment, since the trade was made *in* this PR.

- **`gawk --posix` remains the odd one out** on the single `/dev/stderr` write. Not a regression and not new, but the number of tests riding on that one line keeps growing with each added stderr case; it's the only thing between this script and POSIX-clean.

</details>

<details>
<summary>Nitpicks</summary>

- The header's gap list says a tab-indented fence "reads as indent 0" — true at the top level, but the tab exemption described two bullets earlier means the practical behavior inside a nested fence is the more interesting half. The two bullets are accurate individually; reading them together takes a moment.

</details>

garretpremo (commented):

Approved.

- **The setext gap is the one shape where this branch is strictly worse than `dev`, and that's worth stating plainly even though it shouldn't block.** For a `Title` line over a bare lone `-`, followed by a blank line and a 4-space-indented block whose first line is a fence marker: `dev` returns the reference, this branch returns nothing, silently. CommonMark renders that as an `h2` followed by an *indented* code block, and #129 says indented code stays visible — so this is a real divergence in the direction the header itself argues nothing flags, newly introduced here rather than pre-existing. What makes it non-blocking: the shape is very narrow (a bare lone `-` underline specifically, then 4-space content beginning with a fence marker), the full-repo sweep shows zero real bodies reaching it, the fuzz never generated one, and it is pinned in `documented gaps` with an accurate "HIDES a reference" label plus an explicit `expect(stderr).toBe('')` recording the silence. That's the same way #129 and #130 handled this class. Fixing it needs previous-line state the pre-pass doesn't carry, which is a fair thing to defer — but it deserves a tracking issue rather than only a code comment, since the trade was made *in* this PR.

- **`gawk --posix` remains the odd one out** on the single `/dev/stderr` write. Not a regression and not new, but the number of tests riding on that one line keeps growing with each added stderr case; it's the only thing between this script and POSIX-clean.

PR #149, opened 2026-08-04
fix: only warn on endpoint-less sessionAuth when it carries more than refreshOn (#148)

garretpremo:
Closes #148

## Summary

The `session.endpoint` warning added in #147 was over-broad. It fired whenever `rawSessionAuth` existed without `session.endpoint` — but an endpoint-less `sessionAuth: { refreshOn: [401] }` block is a configuration #135 deliberately supports, since `refreshOn` is sourced from `rawSessionAuth` precisely so it survives the `mergedSessionAuth` narrowing. A correctly-configured project on that route printed the warning on every single CLI invocation.

The warning's actual target is the typo case — a block carrying handshake keys with no reachable endpoint (`sessions:` for `session:`). Gating on "carries some key other than `refreshOn`" keeps that diagnostic and goes quiet for the intentional config.

### `src/auth/refresh-wiring.ts` — gate the warning, and say what was found

```ts
if (rawSessionAuth && !mergedSessionAuth) {
    const foundKeys = Object.keys(rawSessionAuth).filter(
        key => key !== 'refreshOn' && (rawSessionAuth as Record<string, unknown>)[key] !== undefined,
    );

    if (foundKeys.length > 0) {
        console.warn(
            `[apijack] sessionAuth is set but missing session.endpoint — SessionAuthStrategy will not be used. Found: ${foundKeys.join(', ')}.`,
        );
    }
}
```

The `!== undefined` clause is load-bearing, not defensive padding: `deepMergeSessionAuth` always assigns `onChallenge = overrideFn ?? baseFn`, so the key is present-but-`undefined` on every merged block. Without the clause, a refreshOn-only config would still warn.

Naming the found keys is what makes the diagnostic actionable — a `sessions:` typo now reports `Found: sessions, cookies` instead of leaving the user to diff their config against the docs.

### JSDoc no longer claims purity

The function grew an I/O side effect in #147 while its docstring still read "Kept pure so both sites stay in lockstep". The lockstep rationale stands and is kept; the purity claim is replaced with a note about the diagnostic.

## Acceptance criteria

- [ ] A deliberate endpoint-less `sessionAuth: { refreshOn: [...] }` block produces no warning
- [ ] A block with handshake keys but no reachable `session.endpoint` (the typo case) still warns
- [ ] The warning names the offending block's top-level keys
- [ ] `refreshOn` behavior from #135 is unchanged in both cases
- [ ] The `refresh-wiring.ts` JSDoc no longer claims purity

## Test plan

- [x] `bun test` — 1043 pass / 0 fail (up from 1041 on `dev`)

New coverage in `tests/auth/refresh-wiring.test.ts`:

- refreshOn-only block is silent; typo'd block warns and the message names `sessions` and `cookies` but not `refreshOn` (`'sessions'` isn't a substring of `'session.endpoint'`, so that assertion is real rather than incidental)
- both cases still assert `mergedSessionAuth === undefined` **and** `refreshOn === [401]`, so #135's behavior is pinned in both directions
- a third test pins the `onChallenge` case (see below)

Every clause of the new filter was mutation-tested during review — reverting the gate, inverting it, dropping `!== undefined`, or dropping `key !== 'refreshOn'` each fails at least one test, and dropping the `refreshOn` exclusion fails both.

### Known behavior, deliberately kept

`bin/apijack.ts:130-132` and `src/run-routine.ts:140-142` inject `onChallenge` from `.apijack/auth.ts` into the sessionAuth object. A project with a refreshOn-only block **and** a custom `onChallenge` export therefore still warns, reporting `Found: onChallenge` — a key the user never wrote into that block. This is left as-is and pinned by a test: `onChallenge` is consumed only by `SessionAuthStrategy`, which is never constructed without an endpoint, so in that configuration the hook is genuinely dead and the warning is pointing at a real mistake. Excluding it from `foundKeys` would discard a legitimate diagnostic.

### DIFF
--- a/src/auth/refresh-wiring.ts
+++ b/src/auth/refresh-wiring.ts
@@ -18,7 +18,9 @@ export interface RefreshWiring {
 /**
  * Decides the session-auth merge and refresh-retry wiring shared by both
  * cli-builder.ts client-construction sites (the createCli routine-runtime path
- * and the run() path). Kept pure so both sites stay in lockstep (#135).
+ * and the run() path). Shared by both sites so they stay in lockstep (#135);
+ * the only side effect is a diagnostic console.warn on a suspicious sessionAuth
+ * block (see below).
  *
  * `options.refreshOn` (from CliOptions / .apijack/settings.json) takes
  * precedence over `sessionAuth.refreshOn`, so a project can opt a custom
@@ -37,9 +39,20 @@ export function resolveRefreshWiring(
     const refreshOn = options.refreshOn ?? rawSessionAuth?.refreshOn;
 
     if (rawSessionAuth && !mergedSessionAuth) {
-        console.warn(
-            '[apijack] sessionAuth is set but missing session.endpoint — SessionAuthStrategy will not be used.',
+        // An endpoint-less block that carries nothing but `refreshOn` is a deliberate,
+        // supported config (#135) — refreshOn survives the narrowing above precisely so
+        // this works without a session.endpoint. Only warn when there's something else
+        // in the block, which is the signature of a typo'd handshake key (e.g. `sessions:`
+        // instead of `session:`) rather than an intentional refresh-only block.
+        const foundKeys = Object.keys(rawSessionAuth).filter(
+            key => key !== 'refreshOn' && (rawSessionAuth as Record<string, unknown>)[key] !== undefined,
         );
+
+        if (foundKeys.length > 0) {
+            console.warn(
+                `[apijack] sessionAuth is set but missing session.endpoint — SessionAuthStrategy will not be used. Found: ${foundKeys.join(', ')}.`,
+            );
+        }
     }
 
     return { mergedSessionAuth, refreshOn };
--- a/tests/auth/refresh-wiring.test.ts
+++ b/tests/auth/refresh-wiring.test.ts
@@ -50,17 +50,12 @@ describe('resolveRefreshWiring', () => {
         // Not expressible through the SessionAuthConfig type from a fully-typed
         // caller, but envConfig.sessionAuth is only a Partial<SessionAuthConfig> —
         // a JS/dynamic caller could still hand cli-builder a refreshOn-only block.
-        const warnSpy = spyOn(console, 'warn').mockImplementation(() => {});
         const refreshOnlySessionAuth = { refreshOn: [401] } as unknown as SessionAuthConfig;
 
-        try {
-            const result = resolveRefreshWiring({ sessionAuth: refreshOnlySessionAuth }, undefined);
-            expect(result.mergedSessionAuth).toBeUndefined();
-            // refreshOn still surfaces from the raw (unguarded) merge.
-            expect(result.refreshOn).toEqual([401]);
-        } finally {
-            warnSpy.mockRestore();
-        }
+        const result = resolveRefreshWiring({ sessionAuth: refreshOnlySessionAuth }, undefined);
+        expect(result.mergedSessionAuth).toBeUndefined();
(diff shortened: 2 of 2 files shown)


garretpremo (commented):

The gate is correct and every acceptance criterion in #148 is met: the refreshOn-only block goes silent, the typo'd-handshake-key block still warns and now names what it found, `refreshOn` is pinned in both directions, and the JSDoc purity claim is gone. Scope is tight — two files, no drive-by edits — and all 11 checks are green on `55c0284`.

I verified the two load-bearing claims in the description independently:

- `deepMergeSessionAuth` really does always assign `onChallenge` (`src/auth/config-merge.ts:24`, `result.onChallenge = overrideFn ?? baseFn`), and `structuredClone({ ...base, onChallenge: undefined })` preserves the key with an `undefined` value — so `Object.keys` sees `onChallenge` on *every* merged block and the `!== undefined` clause is genuinely required, not padding.
- The `onChallenge` injection at `bin/apijack.ts` and `src/run-routine.ts` is guarded by `if (sessionAuth && projectOnChallenge)`, so it only lands when a project actually exports the hook. Acceptance criterion 1 therefore holds on the real shared-binary path, not just in the unit test.

<details>

- The `Found: onChallenge` case is reasoned about well and pinned by a test, but the message a user actually sees names a key they never typed into their config — they wrote `refreshOn` and a `.apijack/auth.ts` export, and got told `sessionAuth is set but missing session.endpoint ... Found: onChallenge`. The diagnostic is pointing at a real dead hook, so keeping it is the right call; the phrasing just doesn't connect the key back to where it came from. A future tweak could special-case it (`Found: onChallenge (injected from .apijack/auth.ts)`). Not worth blocking on, and not in #148's scope.
- `foundKeys` filters on `!== undefined` only, so a JSON env config with an explicit `"cookies": null` counts as a found key and gets named. That's arguably correct — an explicit null is still something the user wrote — but it's a slightly different notion of "present" than the `undefined` case the filter was written for.
- No caller asserts on the new `Found:` suffix outside the unit test; `tests/cli-builder-refresh-wiring.integration.test.ts` covers the wiring but not the message. Fine as-is given the message is a diagnostic rather than API surface.

</details>

<details>
<summary>Nitpicks</summary>

- The message ends at `Found: sessions, cookies.` without a hint at the fix. For the typo case the found-key list is usually enough to spot `sessions` vs `session`, so this is marginal.
- The description's mutation-testing note is a good signal to have in the PR body — worth keeping that habit, since it's the part of the test suite that's hardest to verify from the diff alone.

</details>

garretpremo (commented):

Approved.

- The `Found: onChallenge` case is reasoned about well and pinned by a test, but the message a user actually sees names a key they never typed into their config — they wrote `refreshOn` and a `.apijack/auth.ts` export, and got told `sessionAuth is set but missing session.endpoint ... Found: onChallenge`. The diagnostic is pointing at a real dead hook, so keeping it is the right call; the phrasing just doesn't connect the key back to where it came from. A future tweak could special-case it (`Found: onChallenge (injected from .apijack/auth.ts)`). Not worth blocking on, and not in #148's scope. (#150)
- `foundKeys` filters on `!== undefined` only, so a JSON env config with an explicit `"cookies": null` counts as a found key and gets named. That's arguably correct — an explicit null is still something the user wrote — but it's a slightly different notion of "present" than the `undefined` case the filter was written for. (#150)
- No caller asserts on the new `Found:` suffix outside the unit test; `tests/cli-builder-refresh-wiring.integration.test.ts` covers the wiring but not the message. Fine as-is given the message is a diagnostic rather than API surface.

PR #156, opened 2026-08-10
fix: read a setext underline as a heading, not a list item (#142)

garretpremo:
Closes #142

## Summary

`scripts/extract-closing-refs.sh`'s awk pre-pass reads a setext H2 underline — a bare lone `-` on the line after paragraph text — as an empty list-item marker, because telling the two apart needs previous-line state the line-oriented pass didn't carry. The container it pushes raises the content column enough that the 4-space-indented block beneath opens as a *fence* (stripped) instead of staying *indented code* (kept, per #129) — silently hiding a real closing reference. A false negative, introduced by #141, tracked by #142.

This PR adds that one bit of look-behind state. The unifying CommonMark rule (every expectation below was verified against the reference `commonmark` parser): **an empty list marker on the line after open paragraph text never starts a list item** — for `-` it is a setext underline (`Title` + `-` → `<h2>`), and for `*`, `+`, `1.`, `1)` it is a lazy paragraph continuation ("an empty list item cannot interrupt a paragraph"). Either way, no container. The one exception, also per CommonMark: a lone marker that dedents back to a sibling position of an open list (`- foo` then `-` at column 0) is a genuine empty sibling item — and in this state machine that case is exactly identifiable as "the line popped a container on its way in".

### `scripts/extract-closing-refs.sh`

One new state variable in the awk pre-pass, `prev_text` — "a paragraph is open":

- Set by any printed non-blank line (including a marker line with content, whose content opens a paragraph); cleared by blank lines, fence openers/content/closers, empty-marker lines, and a line consumed as a setext underline (the underline closes the paragraph, so `Title` / `-` / `-` reads the second `-` as a real empty item — matching CommonMark).
- The pop loop now records whether the line popped a container (`popped`).
- When `list_offset()` sees a marker but the line is an *empty* marker (`^([-*+]|[0-9]+[.)])[[:space:]]*$`), the push is suppressed iff `prev_text && !popped`.

Suppressing a push only ever lowers `base`, which for fence *recognition* strictly errs in the safe direction (a reference stays visible). The final review found the one place that reasoning inverts: `fence_base` also floors the fence *early-close* escape, so a line that sets `prev_text` without opening a CommonMark paragraph (ATX heading, thematic break, HTML block, indented code) can suppress a genuine empty item and let a nested fence swallow a flush-left dedent line — silently. On that shape dev was CommonMark-correct (the reference stays visible prose and the trailing fence line genuinely is unterminated, so dev printed the ref and warned), which makes this a real, accepted trade-off rather than a wash: the shape is contrived (the 150-body sweep never reaches it), so per this file's convention it is **documented as the second known gap that can hide a reference and pinned in `documented gaps`**, not chased. A follow-up issue covers the optional narrowing.

Header comment updated accordingly: the fixed setext behavior is stated where the gap bullet used to be, and the "gaps that can hide a reference" framing is restored to two (early-close reopen, and the new suppression shape).

### `tests/extract-closing-refs.test.ts`

- The old `documented gaps` pin (`a setext underline hides the indented code beneath it — and is silent`) is removed — it pinned the divergence this PR fixes.
- 7 new tests pin the fixed semantics, all pre-verified against reference CommonMark: the underline leaves indented code beneath it visible (`['1']` where dev yields `[]`), trailing-whitespace underline, underline at an item's content column (with a fence proving the no-push: `['7','100']` where dev yields `['7']`), popped sibling still pushes, underline-then-genuine-empty-item, `*`/`1.` cannot interrupt a paragraph, lone `-` after a closed fence still pushes.
- 1 new `documented gaps` pin for the second silent gap above (`[]`, exit 0, empty stderr).

## Acceptance criteria

- [x] A setext H2 underline (bare lone `-` beneath paragraph text) does not push a list container
- [x] The indented code block beneath it stays visible, matching pre-#141 `dev` and reference CommonMark
- [x] Real single-item lists (`-` followed by content, and a lone `-` that genuinely opens an empty list item) still push a container
- [x] The `documented gaps` pin is updated to assert the fixed behavior rather than pinning the divergence
- [x] No regression against the full-repo sweep — dev vs. branch differ on zero real bodies

## Test plan

- [x] `bun test` — 1050 pass / 0 fail (52 in `extract-closing-refs.test.ts`: 7 new semantics tests + 1 new gap pin, 1 stale pin removed)
- [x] `bun run lint` — 0 errors (118 pre-existing warnings, none in changed files)
- [x] Full-repo differential sweep: all 150 issue/PR bodies (47 issues + 103 PRs, all states) run through dev's script and this branch's — 0 stdout/exit-code diffs, 0 stderr diffs
- [x] Every new test expectation cross-checked against the reference `commonmark` npm parser; awk portability re-verified under mawk, busybox awk, nawk, and `gawk --posix`
- [x] Adversarial final review traced the state machine for cross-line leaks (none) and confirmed the fix never hides a reference that dev showed on any constructible input outside the documented gap

### DIFF
--- a/scripts/extract-closing-refs.sh
+++ b/scripts/extract-closing-refs.sh
@@ -46,8 +46,8 @@
 #
 # Known and deliberate gaps. Most leave a reference VISIBLE (a possible false
 # positive) rather than hiding a real one, which is the safer direction to err —
-# the two that can hide one, the early-close reopen above and the setext
-# underline below, say so where they are described:
+# the one that can hide one, the early-close reopen above, says so where it is
+# described:
 #   - indented code blocks and blockquotes are not stripped (per #129: a
 #     blockquote can legitimately carry a real reference)
 #   - code spans are scanned per line, so a span that wraps across a newline
@@ -62,9 +62,6 @@
 #     container early, which only lowers the column a fence must beat
 #   - a fence opener sharing a line with its list marker (`- ```) is not seen,
 #     because the opener is looked for at the line indent, not past the marker
-#   - a setext heading underline (`Title` over a bare `-`) reads as a list item
-#     and pushes a container, since detecting it needs previous-line state; the
-#     container it pushes can open indented code under it as a fence
 #
 # Network-free by design: it takes text, not a PR number. Callers do their own
 # fetching, which differs between them for good reasons (the label scripts hit
@@ -211,7 +208,7 @@ function list_offset(p,   w, n, c, t) {
 
 BEGIN {
     in_fence = 0; fence_char = ""; fence_len = 0; fence_indent = 0; fence_base = 0
-    depth = 0; base = 0
+    depth = 0; base = 0; prev_text = 0
 }
 {
     # Leading spaces, counted with a loop rather than /^ */ because the count is
@@ -242,6 +239,7 @@ BEGIN {
                     fence_indent = 0; fence_base = 0
                 }
             }
+            prev_text = 0
             next
         }
         # Dedented out of the list item holding the fence, so the fence ended
@@ -251,11 +249,13 @@ BEGIN {
         fence_indent = 0; fence_base = 0
     }
 
+    no_para = 0
     if (!blank) {
         # A list item may contain blank lines, so only a non-blank line closes
         # containers. Popping eagerly lowers base, which makes a deep fence
         # LESS likely to be recognized — the false-positive direction.
-        while (depth > 0 && indent < stack[depth]) depth--
+        popped = 0
+        while (depth > 0 && indent < stack[depth]) { depth--; popped = 1 }
         base = 0
         if (depth > 0) base = stack[depth]
 
@@ -267,9 +267,22 @@ BEGIN {
         if (indent <= base + 3) {
             off = list_offset(probe)
             if (off > 0) {
-                depth++
-                stack[depth] = indent + off
-                base = stack[depth]
+                # An empty marker right after open paragraph text never starts
+                # a list item (#142): for `-` it is a setext underline, for the
+                # rest a lazy continuation ("an empty list item cannot
+                # interrupt a paragraph"). The one exception is a marker that
+                # dedented to pop a container on its way in — that is a real
+                # sibling empty item (`- foo` then `-` at column 0), told
+                # apart here by `popped`.
+                empty_marker = (probe ~ /^([-*+]|[0-9]+[.)])[[:space:]]*$/)
+                if (empty_marker && prev_text && !popped) {
+                    no_para = 1
+                } else {
+                    depth++
+                    stack[depth] = indent + off
+                    base = stack[depth]
+                    if (empty_marker) no_para = 1
+                }
             } else if (match(probe, /^(```+|~~~+)/)) {
                 ch = substr(probe, 1, 1)
                 # A backtick info string cannot itself contain a backtick, so
@@ -277,12 +290,14 @@ BEGIN {
                 if (ch != "`" || index(substr(probe, RLENGTH + 1), "`") == 0) {
                     in_fence = 1; fence_char = ch; fence_len = RLENGTH
                     fence_indent = indent; fence_base = base
+                    prev_text = 0
                     next
                 }
             }
         }
     }
     print strip_spans($0)
+    prev_text = (!blank && !no_para)
 }
 
 # A closing fence cannot carry a trailing info string, so one stray character on
--- a/scripts/extract-closing-refs.sh
+++ b/scripts/extract-closing-refs.sh
@@ -44,10 +44,18 @@
 # the state machine is there for exactly that reason. Weigh it before touching
 # the escape.
 #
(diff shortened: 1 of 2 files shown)


garretpremo (commented):

The fix itself is sound: all five acceptance criteria of #142 are met, independent traces of the state machine agree with every new semantics pin, and I reproduced the pinned expectations against the reference `commonmark` parser. What blocks is the recorded justification for the new documented gap — the gap-pin test comment's CommonMark claim is inverted, and that claim is the load-bearing rationale for accepting a new silent false negative.

**Blocking:**
- `tests/extract-closing-refs.test.ts`, the new `documented gaps` pin ("a line that opens no paragraph still suppresses the empty marker under it — silently"): the comment says dev "reports #7 AND fires a spurious unterminated-fence warning, since CommonMark agrees the ref is code." Reference CommonMark says the opposite. It renders `Closes #7` as a **visible paragraph** — the flush-left line ends the list item, and a fenced code block gets no lazy continuation, so the fence closes with the item — and the trailing fence line then opens a genuinely unterminated fence:

  ```html
  <h1>Title</h1>
  <ul><li><pre><code>code
  </code></pre></li></ul>
  <p>Closes #7</p>
  <pre><code></code></pre>
  ```

  So on this shape dev's output (`7` on stdout, unterminated-fence warning on stderr — reproduced against dev's script) was CommonMark-correct on both counts, not "a false positive plus a spurious warning." #157's own Problem section already states the facts correctly ("Reference CommonMark keeps `Closes #7` visible prose"); this comment contradicts it. Accepting the gap is fine per this file's convention — the shape is contrived, it is pinned, and #157 tracks the narrowing — but the pin must record the trade-off honestly: dev showed the reference correctly and warned; this branch hides it silently; accepted because the shape is contrived and the narrowing is scoped to #157. Please reword the test comment (the script's header bullet is accurate and needs no change).

<details>

- Issue #157's "Current state" section repeats the same inverted claim ("dev's symmetric behavior was a false positive plus a spurious unterminated-fence warning rather than correct output"), contradicting its own Problem section two paragraphs up. Worth correcting the issue body alongside the comment fix so the implementer of #157 doesn't inherit the contradiction — likewise the corresponding sentence in this PR's description.
- Suppressed non-`-` empty markers clear `prev_text`, but per CommonMark a lazy continuation is paragraph *text* — the paragraph stays open. On `Title` / `*` / `*` (and `Title` / `*` / `-`), the second lone marker therefore pushes a container here and the indented block beneath opens as a fence, hiding its refs, where reference CommonMark keeps the block indented code (visible). I verified both shapes behave identically on dev (also hidden), so this is not a regression — but if #157 is picked up, clearing `prev_text` only for the `-` suppression (a real underline closes the paragraph) while keeping it set for suppressed `*`/`+`/ordered markers would converge on CommonMark for free.

</details>

<details>
<summary>Nitpicks</summary>

- Same CommonMark-rule family, pre-existing on dev and out of scope here: a non-empty ordered marker other than `1.`/`1)` cannot interrupt a paragraph either (`Title` / `2. foo` is a lazy continuation), but `list_offset` pushes a container for it. Could be one more row in #157.

</details>

garretpremo (commented):

Re-review after the previous "changes requested" round: the single blocking item is resolved. Commit 83eb3d3 rewords the `documented gaps` pin comment exactly as requested — it now records honestly that on that shape dev was CommonMark-correct (printed `#7` and warned) while this branch hides the reference silently, accepted because the shape is contrived and #157 tracks the narrowing. The commit is comment-only; no behavior changed since the round that verified the fix itself, and all five acceptance criteria of #142 remain met. CI is green on the head commit (`83eb3d3`), and I re-ran `tests/extract-closing-refs.test.ts` locally at that commit: 52 pass / 0 fail.

<details>

- Carried over from the previous review, still open and still non-blocking: issue #157's "Current state" section repeats the inverted claim this commit fixed in the test comment, contradicting its own Problem section — worth correcting the issue body (and the corresponding sentence in this PR's description) so the implementer of #157 doesn't inherit the contradiction.
- Also carried over: suppressed non-`-` empty markers clear `prev_text`, though per CommonMark a lazy continuation keeps the paragraph open. Not a regression (dev behaves identically), but if #157 is picked up, clearing `prev_text` only for the `-` suppression would converge on CommonMark for free.

</details>

garretpremo (commented):

Approved.

- Issue #157's "Current state" section repeats the inverted claim this commit fixed in the test comment, contradicting its own Problem section — worth correcting the issue body (and the corresponding sentence in this PR's description) so the implementer of #157 doesn't inherit the contradiction. *(Already addressed: both bodies were corrected via the API shortly before the re-review tick — see the current #157 body and PR description.)*
- Suppressed non-`-` empty markers clear `prev_text`, though per CommonMark a lazy continuation keeps the paragraph open. Not a regression (dev behaves identically), but if #157 is picked up, clearing `prev_text` only for the `-` suppression would converge on CommonMark for free. (#157 — relayed there as a comment.)

PR #158, opened 2026-08-10
fix: validate .apijack/settings.json values instead of casting blindly (#146)

garretpremo:
Closes #146

## Summary

`loadProjectSettings` (`src/settings.ts`) was a bare `JSON.parse` + cast, so a wrong-typed value in `.apijack/settings.json` propagated untouched and failed far from its cause — e.g. `{ "auth": { "refreshOn": 401 } }` (scalar instead of array) reached the generated client and blew up with `this.refreshOn.includes is not a function`.

This PR adds warn-and-ignore validation covering the whole `ProjectSettings` surface: a wrong-typed key emits one warning naming the key and the expected type, and is stripped instead of propagated. A bad settings file never takes the CLI down.

### `src/settings.ts`

Validation applied to the parsed JSON before returning, in place (no whitelist copy — valid keys and unknown keys round-trip identically to before):

- Top-level non-object (array, string, number, `null`, boolean) → warn + return `{}`
- `customCommands` / `customCommands.defaults` / `auth` present but not a plain object → warn + strip
- `customCommands.defaults.requiresAuth` not a boolean → warn + strip
- `auth.refreshOn` not an array of numbers (scalar, or array with a non-number element) → warn + strip
- Nested checks only run when the parent is valid, so one bad container produces exactly one warning
- Presence-gated with `in`, so `requiresAuth: false` and `refreshOn: []` survive validation
- Malformed/unparseable JSON keeps the existing `try/catch` → `{}` behavior

Warning format:

```
[apijack] .apijack/settings.json: 'auth.refreshOn' must be an array of numbers — ignoring
```

### `tests/settings.test.ts`

9 new validation tests (13 total in the file): scalar and mixed-array `refreshOn`, non-boolean `requiresAuth`, the three container-level rules, top-level array/string, valid-file round-trip (including an unknown top-level key) with zero warnings, and one `toBe` assertion pinning the full warning string byte-for-byte.

## Acceptance criteria

- [ ] A wrong-typed value for any `ProjectSettings` key emits a warning naming the key and the expected type, and is ignored rather than propagated
- [ ] A valid settings file behaves exactly as before (no new warnings, unknown keys preserved)
- [ ] A malformed/unparseable file still degrades to `{}` without throwing
- [ ] Test coverage for at least `auth.refreshOn` and `customCommands.defaults.requiresAuth`

## Test plan

- [x] `bun test` — 1053 pass / 0 fail
- [x] `bun run lint` — 0 errors (118 pre-existing warnings, none in the touched files)
- [x] Consumer trace: neither `bin/apijack.ts` nor `src/run-routine.ts` distinguishes a stripped-empty container (`{ auth: {} }`) from an absent one — both read through optional chains with safe fallbacks

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -36,14 +36,14 @@ function validateProjectSettings(parsed: unknown): ProjectSettings {
             warnAndIgnore('customCommands', 'an object');
             delete settings.customCommands;
         } else {
-            const customCommands = settings.customCommands as Record<string, unknown>;
+            const customCommands = settings.customCommands;
 
             if ('defaults' in customCommands) {
                 if (!isPlainObject(customCommands.defaults)) {
                     warnAndIgnore('customCommands.defaults', 'an object');
                     delete customCommands.defaults;
                 } else {
-                    const defaults = customCommands.defaults as Record<string, unknown>;
+                    const defaults = customCommands.defaults;
 
                     if ('requiresAuth' in defaults && typeof defaults.requiresAuth !== 'boolean') {
                         warnAndIgnore('customCommands.defaults.requiresAuth', 'a boolean');
@@ -59,7 +59,7 @@ function validateProjectSettings(parsed: unknown): ProjectSettings {
             warnAndIgnore('auth', 'an object');
             delete settings.auth;
         } else {
-            const auth = settings.auth as Record<string, unknown>;
+            const auth = settings.auth;
 
             if ('refreshOn' in auth) {
                 const refreshOn = auth.refreshOn;
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -14,13 +14,76 @@ export interface ProjectSettings {
     };
 }
 
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+    return typeof value === 'object' && value !== null && !Array.isArray(value);
+}
+
+function warnAndIgnore(key: string, expected: string): void {
+    console.warn(`[apijack] .apijack/settings.json: '${key}' must be ${expected} — ignoring`);
+}
+
+function validateProjectSettings(parsed: unknown): ProjectSettings {
+    if (!isPlainObject(parsed)) {
+        console.warn('[apijack] .apijack/settings.json: settings must be a JSON object — ignoring');
+
+        return {};
+    }
+
+    const settings = parsed as ProjectSettings & Record<string, unknown>;
+
+    if ('customCommands' in settings) {
+        if (!isPlainObject(settings.customCommands)) {
+            warnAndIgnore('customCommands', 'an object');
+            delete settings.customCommands;
+        } else {
+            const customCommands = settings.customCommands as Record<string, unknown>;
+
+            if ('defaults' in customCommands) {
+                if (!isPlainObject(customCommands.defaults)) {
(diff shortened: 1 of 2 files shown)


garretpremo (commented):

All four acceptance criteria from #146 are met and pinned by tests: wrong-typed keys warn with the key path and expected type and are stripped; a valid file (including unknown keys) round-trips with zero warnings; malformed JSON still degrades to `{}`; and both required test cases (`auth.refreshOn`, `customCommands.defaults.requiresAuth`) are covered. Verified in a worktree at the PR head: `bun test tests/settings.test.ts` 13/13 pass, eslint clean on the touched files, `tsc --noEmit` diff-neutral vs base. Consumers read through optional chains (`bin/apijack.ts`, `src/run-routine.ts`), so a stripped key degrades exactly like an absent one. CI is fully green.

<details>

- `src/settings.ts:22` — the warning message hardcodes `.apijack/settings.json` while `loadProjectSettings` takes `apijackDir` as a parameter. Accurate for both current callers, but if a global `~/.<cli>/settings.json` were ever consulted (as `aliases.json` is), the message would point at the wrong file. Interpolating from `settingsPath` would future-proof it.
- Unparseable JSON stays silent (the pre-existing `catch` → `{}`), while a merely wrong-typed value now warns — so a syntax error gets less feedback than a type error. In scope per the criteria (no-throw only), but a one-line warn in the catch would complete the story; candidate follow-up issue.
- Out-of-scope, pre-existing gap worth its own issue: `src/run-routine.ts` never passes `refreshOn: projectSettings.auth?.refreshOn` into its `createCli` call, while `bin/apijack.ts` does — programmatic `runRoutine` consumers don't get the stale-session retry that `settings.json` promises.

</details>

<details>
<summary>Nitpicks</summary>

- Every validation test exercises exactly one invalid key. A case with two invalid keys (e.g. `auth: 'nope'` plus `requiresAuth: 'yes'`) asserting two warnings would pin that validation doesn't short-circuit; likewise `{ "auth": null }` is handled by `isPlainObject` but never pinned.
- The `customCommands` and `auth` blocks share the same nested check-warn-delete shape. Fine at two keys; if `ProjectSettings` grows, a small declarative validator (key path → predicate → expected-type string) would keep warnings uniform.

</details>

garretpremo (commented):

Approved.

- `src/settings.ts:22` — the warning message hardcodes `.apijack/settings.json` while `loadProjectSettings` takes `apijackDir` as a parameter. Accurate for both current callers, but if a global `~/.<cli>/settings.json` were ever consulted (as `aliases.json` is), the message would point at the wrong file. Interpolating from `settingsPath` would future-proof it. (#162)
- Unparseable JSON stays silent (the pre-existing `catch` → `{}`), while a merely wrong-typed value now warns — so a syntax error gets less feedback than a type error. In scope per the criteria (no-throw only), but a one-line warn in the catch would complete the story; candidate follow-up issue. (#162)
- Out-of-scope, pre-existing gap worth its own issue: `src/run-routine.ts` never passes `refreshOn: projectSettings.auth?.refreshOn` into its `createCli` call, while `bin/apijack.ts` does — programmatic `runRoutine` consumers don't get the stale-session retry that `settings.json` promises. (#159)

PR #160, opened 2026-08-10
fix: attribute injected onChallenge in endpoint-less sessionAuth warning (#150)

garretpremo:
Closes #150

## Summary

Polishes the endpoint-less `sessionAuth` diagnostic in `src/auth/refresh-wiring.ts` (spawned from #149's non-blocking review observations). Two items, neither a behavior change to the #148 decision logic:

1. When the framework injects `onChallenge` from `.apijack/auth.ts` into the `sessionAuth` block (which `bin/apijack.ts` and `src/run-routine.ts` both do), the warning now attributes it — the user never wrote that key in their config, so naming it bare pointed them at a key that appears in neither file they'd check.
2. The `foundKeys` filter's treatment of explicit `null` values is now a deliberate, pinned decision: `null` counts as present (the user typed it, and the key is genuinely dead), documented with a comment and covered by a test.

### Attribution plumbing

- `RefreshWiringOptions` (`src/auth/refresh-wiring.ts`) and `CliOptions` (`src/types.ts`) gain an optional `onChallengeInjectedFrom?: string`, documented as framework-set and diagnostic-only.
- Both shared-binary entry points set it to `'.apijack/auth.ts'` in the same block that performs the `sessionAuth.onChallenge = projectOnChallenge` injection — the flag can never be set without the injection having happened. These are the only two production `createCli` call sites.
- No `cli-builder.ts` change: both `resolveRefreshWiring` call sites already pass the whole `options` object, and `CliOptions` is structurally assignable to `RefreshWiringOptions`.

### Warning rendering

When `foundKeys` contains `onChallenge` and the flag is set, the warning renders it with the source; other keys are untouched:

```
[apijack] sessionAuth is set but missing session.endpoint — SessionAuthStrategy will not be used. Found: onChallenge (injected from .apijack/auth.ts), sessions.
```

Rendering is applied to an already-computed key list — the `foundKeys` filter predicate, `refreshOn` precedence, and `mergedSessionAuth` narrowing are byte-for-byte unchanged. A programmatic caller who wrote `onChallenge` into their own `sessionAuth` (no injection) still sees plain `onChallenge`.

The attribution cannot lie: the env-config side of `deepMergeSessionAuth` is JSON-loaded and can never carry a function, so whenever the flag is set, the `onChallenge` being named is the injected one.

### Explicit-null pin

No code change to the filter — a comment on it (referencing #150) records that `value !== undefined` deliberately treats explicit `null` as present, plus a test pinning that `{ cookies: null }` warns and names `cookies`. This holds through the JSON env-config merge path too (`deepMerge` assigns `null` via its scalar branch).

### Tests

Four new tests in `tests/auth/refresh-wiring.test.ts` (existing seven unmodified): injected attribution, non-injected plain rendering, mixed injected + typo'd key (suffix only on `onChallenge`), and the explicit-null pin.

## Acceptance criteria

- [ ] The warning distinguishes a framework-injected `onChallenge` from keys the user wrote into the `sessionAuth` block
- [ ] Explicit-`null` handling in `foundKeys` is decided deliberately and covered by a test
- [ ] The #148 behavior is otherwise unchanged: refreshOn-only is silent, the typo'd-handshake-key case still warns

## Test plan

- [x] `bun test` — 1048 pass / 0 fail (2205 assertions, 83 files)
- [x] `bun run lint` — 0 errors (118 pre-existing warnings in untouched files)
- [x] Independent task review + whole-branch review subagents: clean, no Critical/Important findings
- [x] Verified both `resolveRefreshWiring` call sites pass `options` through, so no cli-builder change is needed

### DIFF
--- a/bin/apijack.ts
+++ b/bin/apijack.ts
@@ -120,6 +120,7 @@ if (!authResolved) {
 
 // 7. Resolve sessionAuth from env config
 let sessionAuth: SessionAuthConfig | undefined;
+let onChallengeInjectedFrom: string | undefined;
 {
     const env = getActiveEnvConfig(CLI_NAME, { configPath: join(configDir, 'config.json') });
 
@@ -129,6 +130,7 @@ let sessionAuth: SessionAuthConfig | undefined;
 
     if (sessionAuth && projectOnChallenge) {
         sessionAuth.onChallenge = projectOnChallenge;
+        onChallengeInjectedFrom = '.apijack/auth.ts';
     }
 }
 
@@ -150,6 +152,7 @@ const cli = createCli({
     specPath,
     auth: authStrategy,
     sessionAuth,
+    onChallengeInjectedFrom,
     refreshOn: projectSettings.auth?.refreshOn,
     generatedDir,
     allowedCidrs: projectConfig?.allowedCidrs,
--- a/src/auth/refresh-wiring.ts
+++ b/src/auth/refresh-wiring.ts
@@ -5,6 +5,10 @@ import { deepMergeSessionAuth } from './config-merge';
 export interface RefreshWiringOptions {
     sessionAuth?: SessionAuthConfig;
     refreshOn?: number[];
+    /** Set by the shared-binary entry points when they inject a project-level
+     *  `onChallenge` (from `.apijack/auth.ts`) into `sessionAuth`. Used only to
+     *  attribute the injected key in the endpoint-less sessionAuth diagnostic. */
+    onChallengeInjectedFrom?: string;
 }
 
 export interface RefreshWiring {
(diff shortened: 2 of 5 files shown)


garretpremo (commented):

All three acceptance criteria from #150 are met and no blocking issues were found. The attribution is pure rendering over an already-computed key list — the `foundKeys` predicate, `refreshOn` precedence, and `mergedSessionAuth` narrowing are byte-for-byte unchanged, and the existing seven #148 tests are unmodified. Verified independently that both `resolveRefreshWiring` call sites in `cli-builder.ts` (L199, L552) pass the whole `options` object, so the new field flows through with no cli-builder change; also verified in `config-merge.ts` that `deepMerge`'s scalar branch assigns explicit `null` (the JSON env-config claim) and that the env-config side can never contribute a function `onChallenge` (the "attribution cannot lie" claim). CI is green on all 12 checks.

<details>

- The explicit-null pin test exercises the filter directly via `options.sessionAuth = { cookies: null }`. The PR body's claim that the JSON env-config merge path preserves `null` (via `deepMerge`'s scalar branch) is correct on code reading, but no test drives `null` through the `envConfig` parameter itself. A one-liner passing `{ sessionAuth: { cookies: null } }` as `envConfig` would pin that route too.
- `onChallengeInjectedFrom` lives on the public `CliOptions`, so a custom-bin consumer could set it and get the `(injected from …)` suffix without any injection having happened. The doc comment marking it framework-set / diagnostic-only is probably sufficient given the field is harmless, but it is now part of the exported type surface.

</details>

garretpremo (commented):

Re-review after 9224205, the only commit since the previous pass at ea95797. It adds exactly the test the prior review's non-blocking observation asked for: `"cookies": null` driven through the `envConfig` parameter of `resolveRefreshWiring`, pinning that `deepMerge`'s scalar branch preserves the explicit `null` and the warning names the key. Verified against `config-merge.ts` that the trace in the test comment is accurate (`null` is falsy, so it falls to the scalar-assignment branch), and that the `onChallenge: undefined` key `deepMergeSessionAuth` always leaves on its result is filtered by the pre-existing `!== undefined` predicate — so the refreshOn-only silence from #148 stays intact on this path too. No production code changed in the new commit; the existing seven #148 tests remain unmodified. All three acceptance criteria from #150 are met and CI is green on all checks.

<details>

- Carried over from the previous review: `onChallengeInjectedFrom` lives on the public `CliOptions`, so a custom-bin consumer could set it and get the `(injected from …)` suffix without any injection having happened. The doc comment marking it framework-set / diagnostic-only is probably sufficient given the field is harmless, but it is part of the exported type surface.

</details>

garretpremo (commented):

Approved.

- Carried over from the previous review: `onChallengeInjectedFrom` lives on the public `CliOptions`, so a custom-bin consumer could set it and get the `(injected from …)` suffix without any injection having happened. The doc comment marking it framework-set / diagnostic-only is probably sufficient given the field is harmless, but it is part of the exported type surface.

Issue #84, opened 2026-04-30
Programmatic `runRoutine()` API + `--json` CLI mode for routines


garretpremo:
## Problem

Routines today are excellent at chaining commands, and `output:` aliases are powerful for inter-step references — but those outputs stay scoped *inside* the routine. There's no documented way to:

1. Invoke a routine from a Node/Bun script and get a typed result back.
2. Run a routine via the CLI and parse its outputs as JSON.

That gap blocks a natural use case: using apijack as a per-test fixture seeder in Playwright (or any test runner). The routine machinery is already doing 95% of what a fixture needs — auth, request chaining, output capture — but a Playwright `beforeEach` has no clean way to pull `output: owner` back out as `{ id, ... }`.

## Proposal

Two surfaces, one engine:

**1. Programmatic API in `@apijack/core`:**

```ts
import { runRoutine } from "@apijack/core";

const result = await runRoutine("seed-user", {
  vars?: Record<string, unknown>,
  env?: string,
  cwd?: string,
});
```

`RoutineResult`:

```ts
type RoutineResult = {
  status: "ok" | "failed";
  output: Record<string, unknown>; // routine's named outputs
  steps: Array<{ name: string; status: "ok" | "failed"; output?: unknown; error?: string }>;
  durationMs: number;
};
```

Existing `output: owner` in YAML surfaces as `result.output.owner`. No YAML changes.

**2. CLI flag:**

```
apijack routine run <name> --json
```

Emits the same `RoutineResult` to stdout. stderr stays silent unless something genuinely failed (logger output suppressed in JSON mode). Exit 0 on `status: "ok"`, non-zero otherwise.

**3. Failure mode (proposed — flag for discussion):** programmatic surface *resolves* with `status: "failed"` rather than throwing, so callers get the partial `steps` array for diagnostics. Throw only on engine-level errors (routine not found, YAML invalid).

**4. Type-safety:** v1 returns `Record<string, unknown>`; callers narrow. Inferred types from YAML output declarations are out of scope.

## Acceptance Criteria

- [ ] `import { runRoutine } from "@apijack/core"` works in a Bun + TypeScript project.
- [ ] `apijack routine run <name> --json` emits parseable JSON to stdout, exit 0 on success / non-zero on failure.
- [ ] Both surfaces share one routine engine — no duplicated step-execution logic.
- [ ] Failure semantics documented: programmatic resolves with `status: "failed"`; CLI exits non-zero.
[...]

PR #163, opened 2026-08-10
fix: fence-aware #N scan, arg validation, pagination rationale in notify-shipped-issues.sh (#140)

garretpremo:
Closes #140

## Summary

Hardens `scripts/notify-shipped-issues.sh` — the release-notification script whose first real run happened with v1.16.0 — per the four acceptance criteria of #140. The `#N` scan is now fence- and inline-code-span-aware, the idempotence guard's pagination rationale matches what the endpoint actually does, argument handling is symmetric (extra positionals now error like unknown flags), and the scan log reads in numeric order.

The fence-awareness decision (criterion 1) is settled by real-world evidence: the v1.16.0 run matched `#14` and `#15` from inline code spans in commit 477bebd's body (`` `prefixes #14` ``, `` `Discloses #15` ``) — they avoided stray "Shipped in" comments only because both happen to be PRs. Implementation reuses the existing stripping stage rather than duplicating it.

### `--bare-refs` mode in `extract-closing-refs.sh`

The shared extractor gains an opt-in flag that keeps the fence/inline-span-stripping pre-pass identical but matches ANY `#[0-9]+` instead of keyword-prefixed closing references. Output stays deduped and numerically ascending. Keyword mode (all three existing callers) is byte-for-byte unchanged — the flag is shifted off before the untouched `case` block, and the keyword pipeline is character-identical.

### `notify-shipped-issues.sh` wiring and fixes

- The scan pipeline `grep -oE '#[0-9]+' | tr -d '#' | sort -u || true` becomes a pipe through `"$SCRIPT_DIR/extract-closing-refs.sh" --bare-refs -`. The breadth (any `#N`, no closing keyword) is unchanged; quoted references inside fenced blocks or inline spans no longer match. Numeric `sort -un` comes with it, fixing the lexicographic log-order nitpick. The old `|| true` is deliberately gone: the extractor exits 0 with empty output for "no refs", so a non-zero exit now means a real internal failure and aborts loudly via `set -e`.
- Log format `"%s %b"` → `"%s%n%b"` so a fence opener on the first body line sits at line start where the stripper can see it.
- Exact-arg-count check: `notify-shipped-issues.sh v1.16.0 <url> junk` now errors with the usage line (exit 1), symmetric with the unknown-flag path.
- The idempotence-guard comment now states what the comments endpoint actually does: ascending-id order, only `since`/`per_page`/`page` supported, so `per_page=100` widens the oldest window rather than biasing toward recent comments; past 100 comments the guard fails open and double-comments (intended direction), with `since`/page-walk named as the exact fix if ever needed.
- Header documents one new known, deliberate gap: the whole range is piped as one document, so an unterminated fence in one commit's body masks references in subsequent commits (stderr warning is the flag). Low probability, missed-courtesy-comment blast radius; per-commit extraction is filed as a follow-up. The stale "Exits 0 unless the arguments themselves are invalid" contract sentence is corrected while touching the header.

### Criterion 4 — first-real-run inspection

Recorded as a comment on #140: the v1.16.0 **Notify shipped issues** step (run 31393124773) was inspected directly — step green, 10 genuinely-shipped closed issues commented, PR references correctly skipped, and 3 prose-cross-reference strays (#84, #91, #98) noted as the documented accepted-breadth tradeoff.

## Acceptance criteria

- [ ] Fence-awareness decided and implemented: `#N` references inside fenced blocks / inline code spans no longer match (decision rationale: real span-sourced false positives observed in the v1.16.0 run)
- [ ] Pagination rationale comment corrected to match the endpoint's actual ascending-id ordering and fail-open-past-100 behavior
- [ ] Nitpicks addressed: extra positional arguments error with usage; scan output numerically sorted
- [ ] Keyword mode of `extract-closing-refs.sh` behaviorally unchanged for its three existing callers
- [ ] First real run inspected directly and recorded on #140 (see issue comment)

## Test plan

- [x] `bun test` — 1076 pass / 0 fail (7 new extractor tests, 4 new notify tests)
- [x] New tests pin: fenced `#N` and inline-span `#N` excluded end-to-end through the notify script (fixture commits with multi-line bodies), fence-opener-as-first-body-line only reachable via `%s%n%b`, numeric scan order (9 → 11 → 100), three-positional usage error, plus extractor-level coverage of bare mode incl. the keyword-mode regression guard
- [x] Independent reviews: per-task spec+quality reviews, whole-branch final review (verdict: ready to merge), scoped re-review of the doc fix wave

### DIFF
--- a/scripts/extract-closing-refs.sh
+++ b/scripts/extract-closing-refs.sh
@@ -81,10 +81,17 @@
 # `gh api repos/.../pulls/N`; collect-closes-refs.sh uses `gh pr view` so it can
 # skip numbers that turn out to be issues rather than PRs).
 #
+# --bare-refs switches the match stage from keyword-prefixed closing
+# references to ANY `#N`, while leaving the code-stripping pre-pass above
+# untouched. Used by scripts/notify-shipped-issues.sh, whose header explains
+# why that breadth — any `#N`, no closing keyword required — is deliberate
+# there; but a `#N` quoted inside code is no more a real reference in bare
+# mode than in keyword mode, so it still goes through the same strip.
+#
 # Usage:
-#   extract-closing-refs.sh --body-file <file>
-#   extract-closing-refs.sh -            # read stdin
-#   cat body.md | extract-closing-refs.sh
+#   extract-closing-refs.sh [--bare-refs] --body-file <file>
+#   extract-closing-refs.sh [--bare-refs] -            # read stdin
+#   cat body.md | extract-closing-refs.sh [--bare-refs]
 #
 # Output: issue numbers, one per line. Empty output and exit 0 when a body has
 # no closing references — that is a normal case (a chore PR opened outside the
@@ -99,6 +106,12 @@
 
 set -euo pipefail
 
+bare_refs=0
+if [ "${1-}" = "--bare-refs" ]; then
+    bare_refs=1
+    shift
+fi
+
 body_file=""
 case "${1-}" in
     --body-file)
@@ -108,7 +121,7 @@ case "${1-}" in
         body_file="/dev/stdin"
         ;;
     *)
-        echo "usage: extract-closing-refs.sh [--body-file <file> | -]" >&2
+        echo "usage: extract-closing-refs.sh [--bare-refs] [--body-file <file> | -]" >&2
         exit 1
         ;;
 esac
@@ -328,8 +341,16 @@ END {
 # rather than be swallowed by the no-match tolerance below.
 stripped=$(awk "$AWK_STRIP" "$body_file")
 
-printf '%s\n' "$stripped" \
-    | grep -oiE "$KEYWORD_RE" \
-    | grep -oE '[0-9]+' \
-    | sort -un \
-    || [ $? -eq 1 ]   # grep exit 1 == no closing references. Any other status propagates.
+if [ "$bare_refs" -eq 1 ]; then
+    printf '%s\n' "$stripped" \
+        | grep -oE '#[0-9]+' \
+        | grep -oE '[0-9]+' \
+        | sort -un \
+        || [ $? -eq 1 ]   # grep exit 1 == no bare references. Any other status propagates.
+else
+    printf '%s\n' "$stripped" \
+        | grep -oiE "$KEYWORD_RE" \
+        | grep -oE '[0-9]+' \
+        | sort -un \
+        || [ $? -eq 1 ]   # grep exit 1 == no closing references. Any other status propagates.
+fi
--- a/scripts/notify-shipped-issues.sh
+++ b/scripts/notify-shipped-issues.sh
@@ -17,7 +17,10 @@
 # This script posts a comment based on the *whole* commit range between
 # releases (any `#N` mention, any keyword), which catches issues referenced
 # in individual commit messages that never made it into the curated PR body.
-# They are complementary — running both is expected, not redundant.
+# They are complementary — running both is expected, not redundant. The
+# `#N` breadth is unchanged, but a reference quoted inside a fenced code
+# block or inline code span is now excluded, via
+# extract-closing-refs.sh --bare-refs.
 #
 # Usage:
 #   notify-shipped-issues.sh <tag> <release-url> [--dry-run]
@@ -35,7 +38,8 @@
 
 set -euo pipefail
 
-source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/gh-pin-account.sh"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/gh-pin-account.sh"
 
 dry_run=0
 args=()
@@ -53,6 +57,11 @@ for arg in "$@"; do
             ;;
     esac
 done
+
+if [ "${#args[@]}" -gt 2 ]; then
+    echo "usage: notify-shipped-issues.sh <tag> <release-url> [--dry-run]" >&2
+    exit 1
+fi
 
 # The `||` chain short-circuits, so `${args[0]}`/`${args[1]}` are never
 # indexed until the length check already confirmed they exist — required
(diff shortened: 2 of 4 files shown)


garretpremo (commented):

All four acceptance criteria of #140 are met and independently verified — nothing blocking. One observation is worth acting on before this ships: the `feat:` prefix on c518ec0 will force a **minor** bump on the next release for a change that never touches the published `@apijack/core` surface.

## Verification performed

Beyond reading the diff, I ran the PR head against real history:

- **Keyword-mode regression (criterion 4 of the PR body).** Differential run of `extract-closing-refs.sh --body-file` on `dev` vs. PR head across the last 400 commit bodies on `main`: **0 output or exit-code differences**. The claim that keyword mode is untouched holds empirically, not just by inspection of the `case` block.
- **Fence/span-stripping (criterion 1).** Swept all 49 stable release ranges (`v0.2.0` → `v1.16.0`), comparing the old `grep -oE '#[0-9]+' | tr -d '#' | sort -u` scan against the new `--bare-refs` pipeline. The *only* difference anywhere in the repo's history is `v1.15.0..v1.16.0`, which now drops `14` and `15` — exactly the two inline-code-span false positives recorded on #140 from commit 477bebd. Nothing genuine is dropped, nothing new is added, and no range emits an `unterminated code fence` warning.
- **Cross-commit fence gap.** Simulated the next release scan (`v1.16.0..` PR head, which includes this PR's own three commits and their fence-heavy bodies): old and new pipelines produce identical output (`138 140 142 144 146 150 152 153 156 158 160`). The documented gap does not bite on the very next release.
- **Tests.** `bun test` on the PR head: 1075 pass, 1 fail — `getPluginPaths() > sourceDir points to project root`, which asserts the checkout path contains `apijack` and fails only because I ran from a worktree at `/tmp/pr163-wt`. It passes in a normal checkout and in CI. Effectively 0 real failures.
- **Lint.** `bun run lint` → 0 errors, 118 pre-existing warnings, matching the PR's test plan.
- **CI.** All 5 checks green (lint, shellcheck, unit-tests on ubuntu + windows, label-new-pr).

## Acceptance criteria

| # | Criterion | Status |
|---|---|---|
| 1 | Fence-awareness decided and implemented | ✅ implemented via `--bare-refs`, decision backed by observed real false positives |
| 2 | Pagination rationale corrected | ✅ new comment states ascending-id order, `since`/`per_page`/`page` only, fail-open past 100 |
| 3 | Nitpicks addressed | ✅ extra-positional check + `sort -un`, both test-pinned |
| 4 | First real run inspected and recorded | ✅ comment on #140 — run 31393124773, step-level inspection, strays enumerated |

<details>

- **`feat:` on c518ec0 will bump the next release to minor.** `scripts/ship.sh:268` matches `^feat[(:]` against commit subjects and sets a minor bump for the whole release. `extract-closing-refs.sh` is release automation — it is not part of the published `@apijack/core` surface, so per the repo's own convention this belongs under `chore:`. It is currently the *only* `feat:` subject in `dev` since `v1.16.0`, so it single-handedly decides minor vs. patch for the next release (and disqualifies the range from `patch-deployer`). Cheapest fix is rewording the subject before merge; otherwise `ship.sh --bump patch` at release time is the escape hatch.
- **`--bare-refs` is only recognized in first position.** `extract-closing-refs.sh --body-file x --bare-refs` silently runs in keyword mode rather than erroring — the trailing arg lands past the `case` block and is dropped. Harmless for the one in-repo caller, and already captured as the "extractor trailing-arg leniency" half of follow-up #164; noting it here only so the two halves stay connected.
- **Cross-commit fence swallowing** is a genuine new way to *hide* a reference, which the extractor's own header calls the dangerous direction. The header documents it, the stderr warning flags it, and #164 tracks the per-commit fix — reasonable to accept. The historical sweep above is the supporting evidence: 49 ranges, zero warnings.

</details>

<details>
<summary>Nitpicks</summary>

- The three new `quotedRepo` tests in `tests/notify-shipped-issues.test.ts` each spawn the script over the same repo and fixture. One run with three assertions would be equivalent and roughly 3x cheaper — though the current shape does read more clearly per-criterion.
- `commitWithMessage` writes `.commit-msg-tmp` inside the fixture repo's working tree. Harmless (every commit is `--allow-empty`, and the file is removed immediately), but `tmpdir()` would keep the fixture repo pristine.
- The PR body's acceptance-criteria checkboxes are all left unchecked despite the work being done.

</details>

garretpremo (commented):

Approved.

- **`feat:` on c518ec0 will bump the next release to minor.** `scripts/ship.sh:268` matches `^feat[(:]` against commit subjects and sets a minor bump for the whole release. `extract-closing-refs.sh` is release automation — it is not part of the published `@apijack/core` surface, so per the repo's own convention this belongs under `chore:`. It is currently the *only* `feat:` subject in `dev` since `v1.16.0`, so it single-handedly decides minor vs. patch for the next release (and disqualifies the range from `patch-deployer`). Cheapest fix is rewording the subject before merge; otherwise `ship.sh --bump patch` at release time is the escape hatch. *(Addressed by decision: left as-is to avoid force-pushing a reviewed branch; release-owner note posted on this PR.)*
- **`--bare-refs` is only recognized in first position.** `extract-closing-refs.sh --body-file x --bare-refs` silently runs in keyword mode rather than erroring — the trailing arg lands past the `case` block and is dropped. Harmless for the one in-repo caller, and already captured as the "extractor trailing-arg leniency" half of follow-up #164. (#164)
- **Cross-commit fence swallowing** is a genuine new way to *hide* a reference, which the extractor's own header calls the dangerous direction. The header documents it, the stderr warning flags it, and #164 tracks the per-commit fix — reasonable to accept. The historical sweep (49 ranges, zero warnings) is the supporting evidence. (#164)

garretpremo (commented):

Approved.

- **`feat:` on c518ec0 will bump the next release to minor.** `scripts/ship.sh:268` matches `^feat[(:]` against commit subjects and sets a minor bump for the whole release. `extract-closing-refs.sh` is release automation — it is not part of the published `@apijack/core` surface, so per the repo's own convention this belongs under `chore:`. It is currently the *only* `feat:` subject in `dev` since `v1.16.0`, so it single-handedly decides minor vs. patch for the next release (and disqualifies the range from `patch-deployer`). Since the merge lands the subject as-is, `ship.sh --bump patch` at release time is the remaining escape hatch.
- **`--bare-refs` is only recognized in first position.** `extract-closing-refs.sh --body-file x --bare-refs` silently runs in keyword mode rather than erroring — the trailing arg lands past the `case` block and is dropped. Harmless for the one in-repo caller, and already captured as the "extractor trailing-arg leniency" half of follow-up #164.
- **Cross-commit fence swallowing** is a genuine new way to *hide* a reference, which the extractor's own header calls the dangerous direction. The header documents it, the stderr warning flags it, and #164 tracks the per-commit fix — reasonable to accept. Supporting evidence from the first-pass sweep: 49 historical release ranges, zero warnings.

PR #165, opened 2026-08-10
fix: forward settings.json auth.refreshOn from runRoutine (#159)

garretpremo:
Closes #159
Closes #161

## Summary

`bin/apijack.ts` forwards `projectSettings.auth?.refreshOn` into `createCli()`. `src/run-routine.ts` loads the same `projectSettings` and forwards `customCommandDefaults` from it — but never `refreshOn`.

The effect: `.apijack/settings.json` with `{ "auth": { "refreshOn": [401] } }` is documented in CLAUDE.md and works from the CLI, but silently does nothing for `runRoutine()`. That's the programmatic entry point CLAUDE.md specifically recommends for Playwright/Vitest fixtures — the long-running flows where a stale session is most likely to bite.

#161 is a duplicate of #159 (filed independently by two workers during the fleet run); both are closed here.

## What changes

One line in `src/run-routine.ts`, mirroring `bin/apijack.ts:156`:

```ts
sessionAuth,
onChallengeInjectedFrom,
refreshOn: projectSettings.auth?.refreshOn,   // ← added
generatedDir,
```

### Test isolation note for the reviewer

The refresh *wiring* below `createCli` is already covered by `cli-builder-refresh-wiring.integration.test.ts`. What was missing — and what let this gap ship — is a test that the programmatic entry point hands the setting down at all. That requires intercepting `createCli`.

A top-level `mock.module` cannot be used here: bun evaluates every test file's top level before running any test, so the stub leaks into the shared module registry. Written that way first, it broke 16 tests across `run-routine.test.ts` and `custom-command-auth.test.ts`. An `afterAll` restore does not help, for the same ordering reason.

So the assertions live in `tests/run-routine-refreshon.isolated.ts` — deliberately named without `.test`, so default discovery skips it — and `tests/run-routine-refreshon.test.ts` runs it as its own `bun test` process, asserting the child's exit code and result line.

## Acceptance criteria

- [ ] `runRoutine()` passes `settings.json` `auth.refreshOn` through to `createCli`
- [ ] An absent `settings.json` leaves `refreshOn` undefined rather than throwing
- [ ] An invalid `auth.refreshOn` is dropped by existing validation before reaching `createCli`
- [ ] No regression in the rest of the suite — specifically no module-mock leakage

## Test plan

- [x] `bun test` — 1077 pass / 0 fail (baseline on this branch: 1076)
- [x] Negative control: with the one-line fix reverted, `auth.refreshOn reaches createCli` fails and the other two still pass — the test is not vacuous
- [x] Leakage check: full suite re-run after isolating the mock into a subprocess returns to 0 fail

### DIFF
--- a/src/run-routine.ts
+++ b/src/run-routine.ts
@@ -159,6 +159,11 @@ export async function runRoutine(
         auth: authStrategy,
         sessionAuth,
         onChallengeInjectedFrom,
+        // Mirror bin/apijack.ts: `.apijack/settings.json` auth.refreshOn must reach
+        // createCli here too. Routine fixtures (Playwright/Vitest) are exactly the
+        // long-running flows a stale session bites, so the programmatic entry point
+        // needs the same stale-session refresh the CLI gets.
+        refreshOn: projectSettings.auth?.refreshOn,
         generatedDir,
         allowedCidrs: projectConfig?.allowedCidrs,
         configPath: join(configDir, 'config.json'),
--- a/tests/run-routine-refreshon.isolated.ts
+++ b/tests/run-routine-refreshon.isolated.ts
@@ -1,0 +1,123 @@
+import { describe, test, expect, beforeEach, afterEach, afterAll, mock } from 'bun:test';
+import { mkdirSync, writeFileSync, rmSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+
+/**
+ * #159 — `runRoutine()` must forward `.apijack/settings.json` `auth.refreshOn`
+ * to createCli(), the same way bin/apijack.ts does.
+ *
+ * The refresh *wiring* below createCli is already covered by
+ * cli-builder-refresh-wiring.integration.test.ts. What was missing — and what
+ * let the gap ship — is a test that the programmatic entry point hands the
+ * setting down at all. So this pins the forwarding, not the retry behavior:
+ * createCli is mocked and we assert on the options object it receives.
+ */
+
+interface CapturedOptions {
+    refreshOn?: number[];
+    customCommandDefaults?: unknown;
+}
+
+let captured: CapturedOptions | undefined;
+
+const stubCli = {
+    use: () => {},
+    command: () => {},
+    dispatcher: () => {},
+    resolver: () => {},
+    runRoutine: async () => ({ status: 'ok' }),
+};
+
+// Capture the real module BEFORE mocking, and put it back in afterAll. bun's
+// module registry is process-wide: without the restore, this stub leaks into
+// every later file that imports createCli (it broke 16 tests across
+// run-routine.test.ts and custom-command-auth.test.ts when first written).
+const realCliBuilder = await import('../src/cli-builder');
+
+mock.module('../src/cli-builder', () => ({
+    ...realCliBuilder,
+    createCli: (options: CapturedOptions) => {
+        captured = options;
+
+        return stubCli;
+    },
+}));
+
+const { runRoutine } = await import('../src/run-routine');
+
+afterAll(() => {
+    mock.module('../src/cli-builder', () => realCliBuilder);
+});
+
+describe('#159 runRoutine forwards settings.json auth.refreshOn', () => {
+    let tmpHome: string;
+    let projectDir: string;
+    let originalHome: string | undefined;
+    let originalCwd: string;
+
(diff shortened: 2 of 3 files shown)


garretpremo (commented):

The one-line fix in `src/run-routine.ts` mirrors `bin/apijack.ts:156` exactly, closing the CLI/programmatic parity gap described in #159 and #161. All acceptance criteria from both issues are met, CI is green, and I reproduced the negative control locally — the new test is not vacuous. No blocking issues.

**Verification performed on this branch:**

- `bun test` — 1077 pass / 0 fail across 84 files (matches the PR's claim; no module-mock leakage)
- `bun run lint` — 0 errors, 118 warnings (all pre-existing)
- Negative control: removed the `refreshOn:` line and re-ran `bun test ./tests/run-routine-refreshon.isolated.ts` → `2 pass / 1 fail`, failing precisely on `auth.refreshOn reaches createCli`. Restored afterward.
- Criterion 3 traced to source: `src/settings.ts:64-69` already rejects a non-`number[]` `auth.refreshOn`, warns, and `delete`s the key — so the invalid case genuinely never reaches `createCli`, rather than being masked by the test.
- The `process.env.HOME` override used by the test is honored by `src/run-routine.ts:43` (`process.env.HOME ?? homedir()`), so the isolation holds on Windows too, where `homedir()` reads `USERPROFILE`. Consistent with the existing convention in `tests/run-routine.test.ts`.

Scope is tight — one production line plus two test files. No public API or generated-output change.

<details>

- `tests/run-routine-refreshon.test.ts` asserts `expect(stderr).toContain('3 pass')`. Adding a fourth assertion to the isolated file will fail the wrapper with a message that points at the wrapper rather than at the real cause. The substring is also loose — `'3 pass'` matches `'13 pass'` and `'23 pass'`. Not worth blocking on, but a future editor of the isolated file will hit it.
- The `afterAll` that restores `../src/cli-builder` in the isolated file is vestigial: the file only ever runs as its own `bun test` process, so nothing later in that registry can observe the stub. The accompanying comment ("without the restore, this stub leaks into every later file that imports createCli") describes the in-process failure mode that motivated the subprocess split, and now reads as if it still applies — it also contradicts the PR description's "an `afterAll` restore does not help, for the same ordering reason." Either drop the `afterAll` and keep the comment purely historical, or reword it to say the restore is belt-and-braces.
- The `.isolated.ts` + spawning-wrapper pattern is new to this repo and is a genuinely reasonable answer to bun's top-level-evaluation ordering. If it is likely to be reused for the next `mock.module`-needing test, a couple of lines in CLAUDE.md's testing notes would save the next author from rediscovering the 16-broken-tests failure mode first.

</details>

<details>
<summary>Nitpicks</summary>

- `expect(`${exitCode} ${exitCode === 0 ? '' : stderr}`.trim()).toBe('0')` encodes diagnostics into an equality assertion. It works and the intent is documented, but `expect(exitCode).toBe(0)` with the stderr logged on failure reads more directly.
- The four-line comment above `refreshOn:` is generous for a line that is a verbatim mirror of `bin/apijack.ts`. Consistent with the `programName` comment just above it, so this is purely taste.

</details>

garretpremo (commented):

Approved.

- `tests/run-routine-refreshon.test.ts` asserts `expect(stderr).toContain('3 pass')`. Adding a fourth assertion to the isolated file will fail the wrapper with a message that points at the wrapper rather than at the real cause. The substring is also loose — `'3 pass'` matches `'13 pass'` and `'23 pass'`. Not worth blocking on, but a future editor of the isolated file will hit it.
- The `afterAll` that restores `../src/cli-builder` in the isolated file is vestigial: the file only ever runs as its own `bun test` process, so nothing later in that registry can observe the stub. The accompanying comment ("without the restore, this stub leaks into every later file that imports createCli") describes the in-process failure mode that motivated the subprocess split, and now reads as if it still applies — it also contradicts the PR description's "an `afterAll` restore does not help, for the same ordering reason." Either drop the `afterAll` and keep the comment purely historical, or reword it to say the restore is belt-and-braces.
- The `.isolated.ts` + spawning-wrapper pattern is new to this repo and is a genuinely reasonable answer to bun's top-level-evaluation ordering. If it is likely to be reused for the next `mock.module`-needing test, a couple of lines in CLAUDE.md's testing notes would save the next author from rediscovering the 16-broken-tests failure mode first.

PR #168, opened 2026-08-11
fix: push dev to origin after the post-release rebase (#167)

### LINKED ISSUE #167: ship.sh never pushes dev after the post-release rebase, leaving origin/dev diverged
## What happens

`scripts/ship.sh` finishes a release by rebasing local `dev` onto `main`, but never pushes `dev`:

```bash
info "Syncing branches..."
git checkout main --quiet && git pull --quiet
git checkout dev --quiet && git rebase main --quiet
```

(`scripts/ship.sh:453-455`, and the identical block at `212-214` inside `resume_untagged_release`.)

So after every release:

- local `main` == `origin/main` ✓
- local `dev` == main's tip (rebased) ✓
- `origin/dev` == still the pre-release commit ✗

## Why it matters

The next `git pull origin dev` refuses with `fatal: Need to specify how to reconcile divergent branches`, because local `dev` and `origin/dev` have diverged. Observed after both v1.16.0 and v1.17.0 — twice in a row, so it is deterministic rather than a one-off.

It also means a fresh clone, a worktree, or an automated agent branching from `origin/dev` starts from the pre-release commit and silently misses the release bump. The parallel issue-fleet branches from `origin/dev`, so this is the state its workers would inherit.

Nothing is lost — the delta is always zero real commits, since the release merged all of `dev` into `main` — but recovering requires a manual `reset --hard origin/dev` or a force-ish reconcile, which is exactly the kind of step that eventually gets done wrong.

## Proposed fix

Push `dev` after the rebase, in both call sites. Never force: if the push is rejected, someone landed work on `dev` during the release window and that work must not be discarded — warn and print the manual reconcile instead.

## Acceptance criteria

- [ ] After a successful ship, `origin/dev`, local `dev`, and `origin/main` all agree
- [ ] The same holds for the `resume_untagged_release` path
- [ ] A rejected push warns with recovery instructions and does not fail the ship — the release has already published by then
- [ ] No force-push under any circumstance

garretpremo:
Closes #167

## Summary

`ship.sh` ended a release by rebasing local `dev` onto `main` and stopping — it never pushed `dev`. So `origin/dev` kept pointing at the pre-release commit while local `dev` sat at main's tip.

Consequences: the next `git pull origin dev` refuses with `fatal: Need to specify how to reconcile divergent branches`, and anything that branches from `origin/dev` — a fresh clone, a worktree, the parallel issue fleet — silently starts from before the release. Observed after both v1.16.0 and v1.17.0, so it was deterministic, not a one-off.

## What changes

The two call sites (the normal path and `resume_untagged_release`) were byte-identical copies — which is precisely why the gap existed in two places. They now share one `sync_branches` helper:

```bash
sync_branches() {
    info "Syncing branches..."
    git checkout main --quiet && git pull --quiet
    git checkout dev --quiet && git rebase main --quiet

    if git push origin dev --quiet 2>/dev/null; then
        ok "dev synced to origin"
    else
        warn "Could not fast-forward origin/dev — someone may have pushed to dev during the release."
        warn "The release itself is fine. Reconcile by hand:"
        warn "  git checkout dev && git pull --rebase origin dev && git push origin dev"
    fi
}
```

**The push is never forced.** A rejected push means someone landed work on `dev` during the release window, and discarding it would be far worse than a stale ref. The helper warns with the manual reconcile and returns success — by the time it runs, the release has already published, so a sync hiccup must not fail the ship.

## Acceptance criteria

- [ ] After a successful ship, `origin/dev`, local `dev`, and `origin/main` all agree
- [ ] Same behavior on the `resume_untagged_release` path
- [ ] A rejected push warns with recovery instructions and does not fail the ship
- [ ] No force-push under any circumstance

## Test plan

New `tests/ship-sync-branches.test.ts` extracts `sync_branches` from `ship.sh` with `sed` and runs it against throwaway `origin.git` + clone pairs, so it exercises the shipped code rather than a reimplementation. Running `ship.sh` end-to-end isn't an option — it merges PRs and publishes to npm.

- [x] Normal post-release: `origin/dev` is fast-forwarded to `main`
- [x] Contended `dev`: push rejected, the other person's commit survives on `origin/dev`, exit code still 0
- [x] Negative control: with the push replaced by `true` (pre-fix behavior), both tests fail
- [x] `bun test` — 1079 pass / 0 fail
- [x] `bash -n scripts/ship.sh` clean; shellcheck runs in CI

## Note

The live divergence from v1.17.0 was already repaired by hand before this PR (`origin/dev` fast-forwarded to `9e78710`, verified as an ancestor first — no force). This change stops it recurring.

### DIFF
--- a/scripts/ship.sh
+++ b/scripts/ship.sh
@@ -84,6 +84,33 @@ promote_issues() {
         || warn "Could not promote issue labels — release has landed; label by hand."
 }
 
+# Sync local branches after a release, and push dev back to origin.
+#
+# The push is the part that used to be missing (#167). Without it, origin/dev
+# stays at the pre-release commit while local dev sits at main's tip: the next
+# `git pull origin dev` refuses with "Need to specify how to reconcile
+# divergent branches", and anything that branches from origin/dev — a fresh
+# clone, a worktree, the parallel issue fleet — silently starts from before
+# the release.
+#
+# Never force. A rejected push means someone landed work on dev during the
+# release window, and that work must not be discarded. Warn with the manual
+# reconcile and carry on: the release has already published by the time this
+# runs, so a sync hiccup must not fail the ship.
+sync_branches() {
+    info "Syncing branches..."
+    git checkout main --quiet && git pull --quiet
+    git checkout dev --quiet && git rebase main --quiet
+
+    if git push origin dev --quiet 2>/dev/null; then
+        ok "dev synced to origin"
+    else
+        warn "Could not fast-forward origin/dev — someone may have pushed to dev during the release."
+        warn "The release itself is fine. Reconcile by hand:"
+        warn "  git checkout dev && git pull --rebase origin dev && git push origin dev"
+    fi
+}
+
 # Resolve a merged PR's merge commit into $MERGE_SHA.
 #
 # `gh pr merge` returns once the REST merge completes, but `gh pr view` reads
@@ -209,9 +236,7 @@ resume_untagged_release() {
     tag_and_publish "$MERGE_SHA" "$main_version"
     promote_issues
 
-    info "Syncing branches..."
-    git checkout main --quiet && git pull --quiet
-    git checkout dev --quiet && git rebase main --quiet
+    sync_branches
     ok "Resumed and shipped v$main_version"
     exit 0
 }
@@ -450,8 +475,6 @@ promote_issues
 
 # ── Step 9: Cleanup ────────────────────────────────────────────────
 
-info "Syncing branches..."
-git checkout main --quiet && git pull --quiet
-git checkout dev --quiet && git rebase main --quiet
+sync_branches
 
 ok "Shipped v$NEW_VERSION"
--- a/tests/ship-sync-branches.test.ts
+++ b/tests/ship-sync-branches.test.ts
@@ -1,0 +1,140 @@
+import { describe, test, expect, beforeEach, afterEach } from 'bun:test';
+import { mkdtempSync, rmSync, writeFileSync } from 'fs';
+import { tmpdir } from 'os';
+import { join } from 'path';
+import { spawnSync } from 'child_process';
+
+const repoRoot = join(import.meta.dir, '..');
+const shipPath = join(repoRoot, 'scripts', 'ship.sh');
+
+/**
+ * #167 — ship.sh's post-release sync must push dev back to origin.
+ *
+ * Before the fix, sync_branches rebased local dev onto main and stopped, so
+ * origin/dev kept pointing at the pre-release commit. The next
+ * `git pull origin dev` refused with "divergent branches", and anything
+ * branching from origin/dev started from before the release.
+ *
+ * sync_branches is extracted from ship.sh rather than reimplemented, so these
+ * tests exercise the shipped code. Running ship.sh end-to-end is not an option
+ * here — it merges PRs and publishes to npm.
+ */
+
+const HELPERS = `
+set -u
+info() { echo "> $1"; }
+ok()   { echo "OK $1"; }
+fail() { echo "FAIL $1"; }
+warn() { echo "WARN $1"; }
+eval "$(sed -n '/^sync_branches() {/,/^}/p' "$SHIP_PATH")"
+`;
+
+interface RunResult {
+    stdout: string;
+    originDev: string;
+    originMain: string;
+    devLog: string;
+}
+
+describe('#167 ship.sh sync_branches pushes dev to origin', () => {
+    let root: string;
+
+    /** origin.git + a clone on `dev`, with `main` one release commit ahead. */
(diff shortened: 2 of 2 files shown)


garretpremo (commented):

All four acceptance criteria from #167 are met, the change is tightly scoped to `scripts/ship.sh` plus one new test file, and CI is green on both `ubuntu-latest` and `windows-latest`. Deduplicating the two byte-identical sync blocks into `sync_branches` is the right shape for a bug whose root cause was the duplication itself.

**Verification performed:**

- **`origin/dev`, local `dev`, `origin/main` agree after a ship** — the release PR is merged with `gh pr merge --merge` (`scripts/ship.sh:459`), so `main`'s second parent is `dev`'s tip. After `git rebase main`, `origin/dev` is an ancestor of local `dev`, making the push a guaranteed fast-forward. Same holds on the resume path, where `main` carries the bump and `dev` has nothing ahead.
- **`resume_untagged_release` path** — `scripts/ship.sh:240` now calls the shared helper; the inline copy is gone.
- **Rejected push warns and does not fail the ship** — the push sits inside an `if`, so `set -e` doesn't trip, and the `else` branch's last command is a `warn`, so the function returns 0. The contended-`dev` test asserts `EXIT=0` explicitly.
- **No force-push** — no `--force`/`--force-with-lease`/`+refs` anywhere in the diff; the contended test asserts the other person's commit survives on `origin/dev`.
- **Negative control reproduced locally** — I extracted `sync_branches` with the same `sed` the test uses, replaced the push with `true`, and ran it against a throwaway `origin.git` + clone pair: `ORIGIN_DEV` and `ORIGIN_MAIN` end up at different SHAs, so test 1's `expect(r.originDev).toBe(r.originMain)` genuinely fails on the pre-fix behavior. The test is not vacuous.
- **Cross-platform** — the new tests are unguarded by `describe.skipIf(process.platform === 'win32')` unlike `extract-closing-refs.test.ts` and `notify-shipped-issues.test.ts`, but the CI log confirms both cases actually executed and passed on `windows-latest` (2.3s each), so no guard is needed.

Extracting the function under test out of `ship.sh` with `sed` rather than reimplementing it is a good call here — it keeps the test honest about the shipped code, which is exactly what a "the code was missing a line" bug needs.

<details>

- `git push origin dev --quiet 2>/dev/null` discards git's stderr. A push that fails for a reason other than contention (network down, expired credentials, a pre-receive hook) still prints "someone may have pushed to dev during the release," which sends the operator looking in the wrong place with no diagnostic to go on. Capturing stderr into a variable and echoing it beneath the warn would keep the message honest without changing control flow.
- The new comment says "a sync hiccup must not fail the ship," but that only holds for the push. `set -euo pipefail` is still in force over the three lines above the `if`: a failed `git pull` on `main`, or a `git rebase main` that hits a conflict, aborts the script after the release has already published — and the rebase case leaves the repo mid-rebase. This is pre-existing behavior, unchanged by the diff and outside #167's scope, but the comment now asserts a robustness property the whole helper doesn't quite have.
- The `resume_untagged_release` call site has no test of its own. It's a one-line substitution of a call for the identical inline block, so inspection is adequate — noting only that acceptance criterion 2 rests on reading the diff rather than on a red/green signal.

</details>

<details>
<summary>Nitpicks</summary>

- `setupRepos()` ignores the `spawnSync` result. If setup ever breaks, the failure surfaces as a puzzling assertion mismatch rather than "setup failed." A `if (r.status !== 0) throw new Error(r.stderr)` would make that a two-second diagnosis.
- `HELPERS` defines `fail()`, which `sync_branches` never calls.
- The `sed -n '/^sync_branches() {/,/^}/p'` extraction depends on the opening line and closing brace staying at column 0. That's true today and any drift fails loudly (undefined function → exit 127 → assertions fail), so it's safe — just a coupling worth knowing about if `ship.sh` is ever reformatted.

</details>

garretpremo (commented):

Approved.

- `git push origin dev --quiet 2>/dev/null` discards git's stderr. A push that fails for a reason other than contention (network down, expired credentials, a pre-receive hook) still prints "someone may have pushed to dev during the release," which sends the operator looking in the wrong place with no diagnostic to go on. Capturing stderr into a variable and echoing it beneath the warn would keep the message honest without changing control flow.
- The new comment says "a sync hiccup must not fail the ship," but that only holds for the push. `set -euo pipefail` is still in force over the three lines above the `if`: a failed `git pull` on `main`, or a `git rebase main` that hits a conflict, aborts the script after the release has already published — and the rebase case leaves the repo mid-rebase. This is pre-existing behavior, unchanged by the diff and outside #167's scope, but the comment now asserts a robustness property the whole helper doesn't quite have.
- The `resume_untagged_release` call site has no test of its own. It's a one-line substitution of a call for the identical inline block, so inspection is adequate — noting only that acceptance criterion 2 rests on reading the diff rather than on a red/green signal.

